PR #33, opened 2021-06-16
Gen867

gyom:
Updated the description of the Mila cluster by explaining a bit how to select which cluster to use.


docs/Information_resources.rst
```diff
@@ -4,4 +4,43 @@ Overview of available computing resources at Mila
 The Mila cluster is to be used for regular development and relatively small
 number of jobs (< 5). It is an heterogeneous cluster. It uses :ref:`SLURM<slurmpage>` to schedule jobs.
 
-.. todo:: Add overview of available resources.
+<<<<<<< HEAD
```
abergeron, line 7: I'm pretty sure this is not supposed to be here.

gyom: Hahaha. Indeed, this is mistake. I'll fix it.

gyom: Fixed.

PR #35, opened 2021-06-16
Improvement for the parallel computation section

abergeron:


achillemyette1:
@abergeron - There is some conflicts here, so I cannot proceed to merge here.

PR #38, opened 2021-06-16
Update requirements and gitignore

breuleux:
There is a small issue where it appears that docutils v0.17 introduces a change in HTML generation that interferes with the style and makes bullet points invisible. The `sphinx_rtd_theme` package seems aware of that incompatibility, so let's use that instead. Also, the build directory should be ignored by git so I added that.

* Add the _build directory to .gitignore
* Use `sphinx_rtd_theme` instead of `sphinx_theme`, which is essentially the same thing but caps docutils to v0.16
* Add a `requirements.txt.lock` file that represents a known working state for the installation


achillemyette1:
@breuleux some conflicts here

PR #70, opened 2021-09-12
Update information about username

btravouillon:
At the time of this commit, over 92% of our user base has their posixUid
identical to their Mila account in our directory database. Some historical
username are indeed made of shorten last name + first letter of first name,
thus it may be worthwhile to keep this information.

### DIFF
--- a/docs/Getting_started.rst
+++ b/docs/Getting_started.rst
@@ -10,8 +10,9 @@ Getting started
 ===============
 
 **Before starting:** get your Mila cluster account ready. Typically, your
-username should look like the first six letters of your last name and the first
-letter of your first name, but this may vary.
+username is identical to your Mila account (the part before ``@mila.quebec``).
+In some special cases, it may look like the first seven letters of your last
+name and the first letter of your first name, but this may vary.
 
 
 Connect to the cluster


docs/Getting_started.rst
```diff
@@ -10,8 +10,9 @@ Getting started
 ===============
 
 **Before starting:** get your Mila cluster account ready. Typically, your
```
btravouillon, line 12: Side note: I  wonder if we should reference the Mila IT Onboarding instead of  `get your Mila cluster account ready`. This would add a link to the Intranet, which is not accessible from the Web. However this could be less ambiguous for newcomers.

@tesfaldet @lavoiems @vict0rsch @mila-iqia/orion any thoughts?

docs/Getting_started.rst
```diff
@@ -10,8 +10,9 @@ Getting started
...
-username should look like the first six letters of your last name and the first
-letter of your first name, but this may vary.
+username is identical to your Mila account (the part before ``@mila.quebec``).
+In some special cases, it may look like the first seven letters of yout last
```
fosterrath-mila: Please review typo : yout -> your.

satyaog (approved):

abergeron (approved):

fosterrath-mila (approved):

docs/Getting_started.rst
btravouillon: #71

fosterrath-mila:
All good thx.

PR #74, opened 2021-10-01
Fix typo in table

btravouillon:
The table of nodes does not print at https://raw.githubusercontent.com/mila-iqia/mila-docs/master/docs/Information_nodes.rst because there is a typo in the table.


fosterrath-mila (approved):

fosterrath-mila:
@satyaog , @achillemyette1 ; Il y a une petite correction à apporter à la doc avec un grand ratio bénéfice/coût. Selon nos conventions, on attend votre approbation.  Merci.

satyaog (approved):
I don't see where the change is. The line looks exactly the same in the GitHub diff viewer but it does fix the issue so I'll approve. However, it would be useful to know what was wrong to avoid making the same mistake

PR #75, opened 2021-10-04
Update documentation with new scratch file system

btravouillon:
The new scratch file system does not support ACLs without enterprise
license, thus I remove the information about ACLs. This could be added
back later if we should invest in enterprise level licenses.

### DIFF
--- a/docs/Userguide_datasets.rst
+++ b/docs/Userguide_datasets.rst
@@ -31,7 +31,7 @@ Mila hosts/seeds some datasets created by the Mila community through `Academic
 Torrent <https://academictorrents.com>`_. The first step is to create `an
 account and a torrent file <https://academictorrents.com/upload.php>`_.
 
-Then drop the dataset in ``/miniscratch/transit_datasets`` and send the
+Then drop the dataset in ``/network/scratch/.transit_datasets`` and send the
 Academic Torrent URL to `Mila's helpdesk <https://it-support.mila.quebec>`_. If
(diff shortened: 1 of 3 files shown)


docs/Userguide_datasets.rst
```diff
@@ -31,7 +31,7 @@ Mila hosts/seeds some datasets created by the Mila community through `Academic
 Torrent <https://academictorrents.com>`_. The first step is to create `an
 account and a torrent file <https://academictorrents.com/upload.php>`_.
 
-Then drop the dataset in ``/miniscratch/transit_datasets`` and send the
+Then drop the dataset in ``/network/scratch/transit_datasets`` and send the
```
btravouillon, line 34: Should we hide the directory?  `/network/scratch/.transit_datasets`
It has a specific purpose which is defined here. A copy & paste of the path should be ok.

btravouillon:
Draft: don't merge until Wednesday 6 October 10:00 AM.

docs/Userguide_datasets.rst
satyaog: I believe we chose the hidden approach right (`/network/scratch/.transit_datasets`)? I'm ok with both

btravouillon: Indeed, I updated the change to reflect our decision.

fosterrath-mila (approved):
Merci Bruno!

achillemyette1 (approved):

Issue #51, opened 2021-08-17
Provide a link to some SSH tunnelling guide


satyaog:
![Screen Shot 2021-08-07 at 3 44 27 PM](https://user-images.githubusercontent.com/5014902/128612175-d13dbfc7-233b-4988-8a3e-eb2feefec9ef.png)
Change to "only when using the Mila wifi or through SSH tunnelling". Provide a link to some SSH tunnelling guide.

_Originally posted by @tesfaldet in https://github.com/mila-iqia/mila-docs/issues/46#issuecomment-894699291_

btravouillon:
We can add an example like `ssh -L19999:<node>.server.mila.quebec:19999 -p 2222 login.server.mila.quebec` and a link to http://localhost:19999.
Anyone who want to know more about "SSH tunneling" should be able to use their favorite search engine. (URLs on the Internet change every now and then, we can't deal with this in the long run).

fosterrath-mila:
Peut-on matérialiser le commentaire de Bruno dans un PR ?

satyaog:
Closed as this was covered in #76

Issue #16, opened 2021-05-22
Add section on remote SSH through VS Code


tesfaldet:
There are plenty of students wanting to remotely develop on the cluster using VS Code's Remote SSH extension. The reason being that it offers remote debugging capabilities in a flexible IDE, which can prove invaluable for rapid prototyping. I suggest a section to be added on using VS Code with the cluster. I can contribute the first PR if need be.

tesfaldet:
Below is a copy-paste of my explanation of getting this working with a student on Slack. Mind the weird sentence structuring...

Save this file in your home directory `$HOME/dev_env.sh`

```
#!/bin/bash
#SBATCH --job-name=dev_env
#SBATCH --ntasks=1
#SBATCH --mem=12Gb
#SBATCH --cpus-per-task=2
#SBATCH --gres=gpu:1
#SBATCH --output=%x_%A.out
#SBATCH --error=%x_%A.err
​
tail -f /dev/null
```

To use this, first you have to make the script executable. `chmod +x $HOME/dev_env.sh`. Then, run `sbatch $HOME/dev_env.sh` from a login node (`ssh -p 2222 <user>@login.server.mila.quebec`) This will request a job with the bash code `tail -f /dev/null` to be executed by a compute node. What this does is make the job never stop until it is pre-empted, but don’t worry it’s not a busy wait so it’s not wasting any resources. Once the job starts, you can connect to its node from within the login node or even locally on your home machine. To connect from within the login node, `ssh <compute node hostname>`. To connect locally from home, `ssh -J <user>@login.server.mila.quebec:2222 <user>@<compute node hostname>`.

To connect locally from home but through VS Code’s Remote SSH extension, add this in your local ssh config file (`~/.ssh/config`)

```
Host mila_login
    HostName login.server.mila.quebec
    User <user>
    Port 2222
    PreferredAuthentications publickey,keyboard-interactive
    ServerAliveInterval 120
    ServerAliveCountMax 5
Host mila_compute
    HostName <compute node hostname>
    User <user>
    ProxyJump mila_login
```

Then using the Remote SSH VS Code extension, connect to the host `mila_compute`

Also, if you have this ssh config file setup this way, you can also do `ssh mila_compute` instead of `ssh -J <user>@login.server.mila.quebec:2222 <user>@<compute node hostname>` if you want to connect via your terminal.

Oh, and to get the compute node hostname after you’ve launched your job, run `squeue -u $USER` in the login node.

```
[...]

PR #76, opened 2021-10-07
Update login guide, add section on VSCode

breuleux:
Tell me if I should split this into multiple PRs.

* Update the user guide with information from the Getting Started guide.
* Deactivate the Getting Started guide because of redundancies with the rest of the docs (it is now a link to the user guide).
* Add a page on Visual Studio Code, including a link to milatools and some of @tesfaldet's tips (#16)
* Address #51

### DIFF
--- a/docs/Userguide_login.rst
+++ b/docs/Userguide_login.rst
@@ -57,10 +57,9 @@ mila init
 ---------
 
 To make it easier to set up a productive environment, Mila publishes the
-milatools_ package, which defines
-a ``mila init`` command which will automatically perform some of the below steps
-for you. You can install it with ``pip`` and use it, provided your Python
-version is at least 3.8:
+milatools_ package, which defines a ``mila init`` command which will
+automatically perform some of the below steps for you. You can install it with
+``pip`` and use it, provided your Python version is at least 3.8:
 
 .. code-block:: bash
 
--- a/docs/VSCode.rst
+++ b/docs/VSCode.rst
@@ -4,32 +4,36 @@ Visual Studio Code
 
 One editor of choice for many researchers is VSCode. One feature of VSCode is
 remote editing through SSH. This allows you to edit files on the cluster as if
-they were local. You can also debug your programs using VSCode's debugger,
-open terminal sessions, etc.
+they were local. You can also debug your programs using VSCode's debugger, open
+terminal sessions, etc.
 
 
 Connecting to the cluster
 -------------------------
 
 VSCode cannot be used to edit code on the login nodes, because it is a heavy
-enough process (a ``node`` process, plus the language server, linter, and possibly
-other plugins depending on your configured environment) that there is a risk of
-overloading the login nodes if too many researchers did it at the same time.
+enough process (a ``node`` process, plus the language server, linter, and
+possibly other plugins depending on your configured environment) that there is a
+risk of overloading the login nodes if too many researchers did it at the same
+time.
 
-Therefore, to use VSCode on the cluster, you first need to allocate a compute node,
-then connect to that node.
+Therefore, to use VSCode on the cluster, you first need to allocate a compute
+node, then connect to that node.
 
-The milatools_ package provides a command to make the operation easier. Simply run:
+The milatools_ package provides a command to make the operation easier. Simply
+run:
 
 .. code-block:: bash
 
     mila code path/to/code
 
-Note that the above command requires your SSH config to be organized in a certain way,
-so you should run ``mila init`` prior to using ``mila code`` for the first time.
+Note that the above command requires your SSH config to be organized in a
+certain way, so you should run ``mila init`` prior to using ``mila code`` for
+the first time.
 
-If you already have an allocation, or you want to run VSCode on the same node as one
-of your running jobs, the ``--node name`` and ``--job jobid`` options can be used.
+If you already have an allocation, or you want to run VSCode on the same node as
+one of your running jobs, the ``--node name`` and ``--job jobid`` options can be
+used.
 
 .. _milatools: https://github.com/mila-iqia/milatools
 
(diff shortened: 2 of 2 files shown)


docs/Information_monitoring.rst
```diff
@@ -7,6 +7,9 @@ This information is exposed in two ways:
 
 * For every node, there is a web interface from Netdata itself at ``<node>.server.mila.quebec:19999``.
   This is accessible only when using the Mila wifi or through SSH tunnelling.
+
+  * SSH tunnelling: on your local machine, run ``ssh -L19999:<node>.server.mila.quebec:19999 -p 2222 login.server.mila.quebec``,
+    then open http://localhost:19999 in your browser.
```
satyaog, line 12: Should we replace the port 19999 by `<PORT>`? I fear that many will end up using the 19999 port on the same node which could potentially bring some issue

docs/Userguide_login.rst
```diff
@@ -13,34 +13,139 @@ You can access the Mila cluster via ssh:
...
+    SHA256:baEGIa311fhnxBWsIZJ/zYhq2WfCttwyHRKzAb8zlp8 (ECDSA)
+    SHA256:XvukABPjV75guEgJX1rNxlDlaEg+IqQzUnPiGJ4VRMM (DSA)
+    SHA256:Xr0/JqV/+5DNguPfiN5hb8rSG+nBAcfVCJoSyrR0W0o (RSA)
+    SHA256:gfXZzaPiaYHcrPqzHvBi6v+BWRS/lXOS/zAjOKeoBJg (ED25519)
+
+
+.. important::
+    Login nodes are merely *entry points* to the cluster. They give you access
+    to the compute nodes and to the filesystem, but they are not meant to run
+    anything heavy. Do **not** run compute-heavy programs on these nodes,
```
satyaog, line 38: I think I would say _You should not run anything different then a text editor because otherwise you could bring them down, impeding cluster access for everyone._. But it might be to aggressive

satyaog (approved):
Some minor stuff that could be changed but not mandatory I believe. And very impressive and useful app that is the `milatools`

docs/Getting_started.rst
```diff
@@ -1,261 +1,270 @@
...
+Getting started
+===============
 
-.. It seems this section would need to be broken up into the current existing
-  sections (cluster theory and / or user_guide), hence I am not reactivating this
-  section just yet.
+See :ref:`userguide`.
 
 
+.. **Before starting:** get your Mila cluster account ready. Typically, your
```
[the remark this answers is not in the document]
breuleux: Possibly, in the later sections. This being said, it's still in git history so might as well erase it, yeah.

docs/index.rst
```diff
@@ -31,7 +31,6 @@ Welcome to Mila's technical documentation. See contents below.
    :caption: How-tos and Guides
    :maxdepth: 2
 
-   Getting_started
```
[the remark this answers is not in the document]
breuleux: I was told that the redundancy confused some people, who were not sure what the intended difference was between the two pages.

docs/Information_monitoring.rst
breuleux: It has to be port `19999` on the compute node because that is where Netdata is serving the data. There's only going to be an issue if a user wants to monitor two different nodes on the same machine, in which case they'd have to use different local ports, but I don't know if it's worth mentioning, it's not super hard to figure out.

docs/Userguide_login.rst
breuleux: I mean, we don't want them to run VSCode on the login nodes, so even that might send mixed messages ;)

docs/VSCode.rst
```diff
@@ -4,12 +4,74 @@ Visual Studio Code
...
+    interpreter selection prompt in VSCode to use that same environment.
+
+
+Troubleshooting
+---------------
+
+"Cannot reconnect"
+^^^^^^^^^^^^^^^^^^
+
+When connecting to multiple compute nodes (and/or from multiple computers), some instances may crash with that message because of conflicts in the lock files VSCode installs in ``~/.vscode-server`` (which is shared on all compute nodes through NFS). To fix this issue, you can change this setting in your ``settings.json`` file:
```
btravouillon, line 61: La plupart du contenu wrap à 80 colonnes.

btravouillon, line 61: `/home/mila` is not NFS.

```suggestion
When connecting to multiple compute nodes (and/or from multiple computers), some instances may crash with that message because of conflicts in the lock files VSCode installs in ``~/.vscode-server`` (which is shared on all compute nodes). To fix this issue, you can change this setting in your ``settings.json`` file:
```

docs/VSCode.rst
```diff
@@ -4,12 +4,74 @@ Visual Studio Code
...
+
+.. code-block:: json
+
+    "remote.SSH.lockfilesInTmp": true
+
+This will store the necessary lockfiles in ``/tmp`` on the compute nodes (which are not shared through NFS).
```
btravouillon, line 67: Open question: can we use `$SLURM_TMPDIR` instead of `/tmp` for lockfiles?

btravouillon, line 67:
```suggestion
This will store the necessary lockfiles in ``/tmp`` on the compute nodes (which are local to the node).
```

docs/Information_monitoring.rst
btravouillon: I think it's clear enough. If one try to open a connection to a different port with the same local port, ssh will display:
```
bind [127.0.0.1]:19999: Address already in use
channel_setup_fwd_listener_tcpip: cannot listen to port: 19999
Could not request local forwarding.
```

docs/Userguide_login.rst
satyaog: True even if VSCode is more of an IDE than a text editor but users might not do the difference

docs/index.rst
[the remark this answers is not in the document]
breuleux: I feel it might still be relevant, but we need better planning and integration with the rest of the docs.

docs/VSCode.rst
tesfaldet: This is a common setup for development environments that house a single vs code install at a shared directory used by multiple users (or multiple vs code instances) at the same time. However, yeah, it could break stuff if you're unlucky I think. Probably requires some more thought behind it but it does solve a problem for now, which is having vs code open from multiple computers at the same time (e.g., my laptop and my desktop).

docs/index.rst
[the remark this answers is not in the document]
satyaog: @fosterrath-mila @achillemyette1 ?

PR #96, opened 2021-12-02
Add Slurm+HPO example

Delaunay:


docs/examples/hpo_launcher.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Usage:
+#
+#   sbatch --array=0-100 --gres=gpu:1 --cpus-per-gpu=2 --mem-per-gpu=16G scripts/hpo_launcher.sh train.py
+#
+
+# Slurm configuration
+# ===================
+
+#SBATCH --exclude=kepler4,kepler3
```
btravouillon, line 12: Is it possible to remove this very specific exclude?
Would it make more sense to specify `--gres=gpu:rtx8000:1` as an example instead?

docs/Userguide_HPO.rst
```diff
@@ -0,0 +1,110 @@
...
+
+This means there could be 400 set of hyperparameters being worked on in parallel across 100 GPUs.
+
+The easiest way to run an hyperparameter search on the cluster is simply to use a job array,
+which will launch the same job n times. Your HPO library will generate different parameters to try
+for each instances.
+
+.. code-block
+
+    sbatch --array=0-100 --gres=gpu:1 --cpus-per-gpu=2 --mem-per-gpu=16G scripts/hpo_launcher.sh train.py
```
btravouillon, line 20: `--array=0-100` means 101 tasks, isn't it?

docs/Userguide_HPO.rst
```diff
@@ -0,0 +1,110 @@
...
+
+
+.. code-block:: bash
+
+   orion hunt --config $ORION_CONFIG python train.py --config $SPACE_CONFIG --arg1 value1
+
+
+.. note::
+
+    Do not forget to move orion database out of ``scratch`` once the experiment is done.
```
btravouillon, line 92:
```suggestion
   Do not forget to move orion database out of ``scratch`` once the experiment is done.
```

docs/examples/hpo_launcher.sh
```diff
@@ -0,0 +1,73 @@
...
+
+conda activate py39
+
+# Environment
+# ===================
+
+export EXPERIMENT_NAME="MyExperiment"
+
+# Constant
+export SCRATCH=~/scratch
```
btravouillon, line 25: `$SCRATCH` is now defined in the Mila cluster.

satyaog (commented):
This looks good but my only question would be on the compute intensity of the example. How much compute does the 100 jobs (or 101) require? I would think we would not want to observe a lot more than 0.5 GPU/h and 1 CPU/h to check how a concept is working in practice. Thoughts?

ahmam (approved):

docs/Userguide_HPO.rst
Delaunay: ah right

Delaunay: fixed

docs/examples/hpo_launcher.sh
Delaunay: fixed

Delaunay:
> This looks good but my only question would be on the compute intensity of the example. How much compute does the 100 jobs (or 101) require? I would think we would not want to observe a lot more than 0.5 GPU/h and 1 CPU/h to check how a concept is working in practice. Thoughts?


100 trials with 10 epochs on cifar  is actually quite fast = (1 trial with 1000 epochs)!
It also show case how easy it is to do HPO and why they should do it.

docs/examples/hpo_launcher.sh
btravouillon: `kepler[3,4]` are still in production. They should be removed this year however.

docs/Userguide_HPO.rst
```diff
@@ -0,0 +1,110 @@
...
+`Orion <https://orion.readthedocs.io/en/stable/?badge=stable>`_
+is an asynchronous framework for black-box function optimization developped at Mila.
+
+Its purpose is to serve as a meta-optimizer for machine learning models and training,
+as well as a flexible experimentation platform for large scale asynchronous optimization procedures.
+
+Orion saves all the results of its optimization process in a database,
+by default it is using a local database on a shared filesystem named ``pickleddb``.
+You will need to specify its location and the name of your experiment.
+Optionally you can configure workers which will run on parallel to maximize resource usage.
```
btravouillon, line 34:
```suggestion
Optionally you can configure workers which will run in parallel to maximize resource usage.
```

docs/Userguide_HPO.rst
```diff
@@ -0,0 +1,110 @@
+Hyperparameter Optimization
+===========================
+
+Hyperparameter optimization is very easy to parallelize as each trial (unique set of hyperparameters) are
+independant of each other. The easiest way is to launch as many jobs as possible each trying a different
+set of hyperparameters and reporting their results back to a synchronized location (database).
+
+You will need to estimate how much resources your training requires and update the provided example to fit.
+In the example below we use 100 tasks with each 4 CPU cores and one GPU.
+Each task will run 4 training in parallel on the same GPU to maximize its utilization.
```
btravouillon, line 10:
```suggestion
Each task will run 4 trainings in parallel on the same GPU to maximize its utilization.
```

docs/Userguide_HPO.rst
```diff
@@ -0,0 +1,110 @@
...
+         "lr": "orion~loguniform(1e-5, 1.0)",
+         "weight_decay": "orion~loguniform(1e-10, 1e-3)",
+         "momentum": "orion~loguniform(0.9, 1.0)"
+      }
+   EOM
+
+Run Orion hunt
+^^^^^^^^^^^^^^
+
+Now that orion is configured and has a search space we can execute orion-hunt
```
btravouillon, line 81:
```suggestion
Now that Orion is configured and has a search space we can execute orion-hunt
```
It is noted **Orion** elsewhere in the file.

docs/examples/hpo_launcher.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+set -v
+
+# Usage:
+#
+#   sbatch --array=0-100 --gres=gpu:1 --cpus-per-gpu=2 --mem-per-gpu=16G scripts/hpo_launcher.sh train.py
```
btravouillon, line 6:
```suggestion
#   sbatch --array=1-100 --gres=gpu:1 --cpus-per-gpu=2 --mem-per-gpu=16G scripts/hpo_launcher.sh train.py
```

btravouillon (commented):
Some comments in-line.

I also wonder why do we find **hyperparameter**, **hyper parameter** and **hyper-parameter**. Is there a different wording for different cases?

Delaunay:
> Some comments in-line.
> 
> I also wonder why do we find **hyperparameter**, **hyper parameter** and **hyper-parameter**. Is there a different wording for different cases?

@btravouillon 

I harmonized the naming

btravouillon (approved):
Looks good to me. 👍
Merci !

achillemyette1 (approved):

PR #120, opened 2022-05-27
Remove DSA fingerprint for login nodes

btravouillon:
The login nodes do not use the DSA key anymore.


obilaniu (approved):
Full Approval. DSA is not recommended anymore and shouldn't be part of the cypher suite.

achillemyette1 (approved):

Issue #115, opened 2022-04-19
Add howto on data sharing


fosterrath-mila:
The sharing data section in the policy section is very limited. 
There is a need for new users to have a practical section on HOW to share data.

Write up a practical guide giving examples to attain data sharing goals.

btravouillon:
Closed by #121.

PR #121, opened 2022-06-01
Add ACL example

satyaog:
Relates to #115

### DIFF
--- a/docs/Userguide_sharing_data.rst
+++ b/docs/Userguide_sharing_data.rst
@@ -66,8 +66,8 @@ search and delete each others' files:
    setfacl -Rm  user:$USER2:rwx /network/scratch/${USER:0:1}/$USER/X/Y/Z/
 
 .. note::
-   The purpose of granting permissions first for _future_ files and then for
-   _existing_ files is to prevent a **race condition** whereby after the first
+   The purpose of granting permissions first for *future* files and then for
+   *existing* files is to prevent a **race condition** whereby after the first
    ``setfacl`` command the other person could create files to which the
    second ``setfacl`` command does not apply.
 
--- a/docs/Userguide_sharing_data.rst
+++ b/docs/Userguide_sharing_data.rst
@@ -1,0 +1,101 @@
+Sharing Data with ACLs
+======================
+
+As an illustrative example, to allow ``$USER`` to share with ``$USER2`` in
+``rwx`` a hierarchy ``/network/scratch/$USER/X/Y/Z/...`` with ACLs:
+
+----
+
+| Grant **oneself** permissions to access any **future** files/folders created
+  by the other *(or oneself)*
+| (``-d`` renders this permission a "default" / inheritable one)
+
+.. code-block:: bash
+
+   setfacl -Rdm user:$USER:rwx  /network/scratch/$USER/X/Y/Z/
+
+----
+
+| Grant **another** permission to access any **future** files/folders created
+  by the other *(or oneself)*
+| (``-d`` renders this permission a "default" / inheritable one)
+
+.. code-block:: bash
+
+   setfacl -Rdm user:$USER2:rwx /network/scratch/$USER/X/Y/Z/
+
+----
+
+| Grant **another** permission to access any **existing** files/folders created
+  by *oneself*.
+| Such files and folders were created before the new default ACLs were added
+  above and thus did not inherit them from their parent folder at the moment of
+  their creation.
+
+.. code-block:: bash
+
+   setfacl -Rm  user:$USER2:rwx /network/scratch/$USER/X/Y/Z/
+
+----
+
+| Grant **another** permission to search through one's hierarchy down to the
+  shared location in question.
+
+* **Non**-recursive (!!!!)
+* May also grant ``:rx`` in unlikely event others listing your folders on the
+  path is not troublesome or desirable.
+
+----
+
+| In order to access a file, all folders from the root (``/``) down to the
+  parent folder in question must be searchable (``+x``) by the concerned user.
+  This is already the case for all users for folders such as ``/``, `/network`
+  and ``/network/scratch``, but users must explicitly grant access to some or
+  all users by adding ACLs for ``/network/scratch/$USER``, ``$HOME`` and
+  subfolders.
+
+.. code-block:: bash
+
+   setfacl -m   user:$USER2:x   /network/scratch/$USER/X/Y/
+   setfacl -m   user:$USER2:x   /network/scratch/$USER/X/
+   setfacl -m   user:$USER2:x   /network/scratch/$USER/
+
+----
+
+| To bluntly allow **all** users to search through a folder (**think twice!**),
+  the following command can be used:
+
+.. code-block:: bash
+
+        chmod a+x /network/scratch/$USER/
+
+----
+
+.. note::
+  * ``man setfacl``
+  * ``man path_resolution``
+
+Viewing and Verifying ACLs
+--------------------------
+
+.. code-block:: bash
+
+   getfacl /path/to/folder/or/file
+              1:  # file: somedir/
+              2:  # owner: lisa
+              3:  # group: staff
+              4:  # flags: -s-
+              5:  user::rwx
+              6:  user:joe:rwx               #effective:r-x
+              7:  group::rwx                 #effective:r-x
+              8:  group:cool:r-x
+              9:  mask::r-x
+             10:  other::r-x
+             11:  default:user::rwx
+             12:  default:user:joe:rwx       #effective:r-x
+             13:  default:group::r-x
+             14:  default:mask::r-x
+             15:  default:other::---
+
+.. note::
+  * ``man getfacl``
--- a/docs/Userguide_sharing_data.rst
+++ b/docs/Userguide_sharing_data.rst
@@ -1,32 +1,61 @@
 Sharing Data with ACLs
 ======================
 
-As an illustrative example, to allow ``$USER`` to share with ``$USER2`` in
-``rwx`` a hierarchy ``/network/scratch/$USER/X/Y/Z/...`` with ACLs:
+Regular permissions bits are extremely blunt tools: They control access through
+only three sets of bits owning user, owning group and all others. Therefore,
+access is either too narrow (``0700`` allows access only by oneself) or too wide
+(``770`` gives all permissions to everyone in the same group, and ``777`` to
+literally everyone).
+
+ACLs (Access Control Lists) are an expansion of the permissions bits that allow
+more fine-grained, granular control of accesses to a file. They can be used to
+permit specific users access to files and folders even if conservative default
+permissions would have denied them such access.
+
+
+As an illustrative example, to use ACLs to allow ``$USER`` (**oneself**) to
+share with ``$USER2`` (**another person**) a "playground" folder hierarchy in
+Mila's scratch filesystem at a location
+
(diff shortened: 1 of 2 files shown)


satyaog:
https://mila-docs-staging--121.org.readthedocs.build/en/121/Userguide.html#sharing-data-with-acls

satyaog:
@obilaniu , could you quickly look if this is the format you had in mind?

obilaniu (approved):
It's all good! Thanks for translating from Markdown!

docs/Userguide_sharing_data.rst
```diff
@@ -0,0 +1,101 @@
...
+
+* **Non**-recursive (!!!!)
+* May also grant ``:rx`` in unlikely event others listing your folders on the
+  path is not troublesome or desirable.
+
+----
+
+| In order to access a file, all folders from the root (``/``) down to the
+  parent folder in question must be searchable (``+x``) by the concerned user.
+  This is already the case for all users for folders such as ``/``, `/network`
```
btravouillon, line 52:
```suggestion
  This is already the case for all users for folders such as ``/``, ``/network``
```

btravouillon (changes requested):
One tiny typo.

Also, while this is very cool, this does not cover all topics of #115. I believe we won't be able to address all topics in a single PR, so I will remove the `resolves #115` to avoid case closure.

btravouillon:
@vict0rsch: Any feedback on these ACL examples?

docs/Userguide_sharing_data.rst
```diff
@@ -0,0 +1,101 @@
+Sharing Data with ACLs
+======================
```
vict0rsch, line 2: Maybe start by explaining what ACLs are, _and_ provide a link to learn (like https://www.thegeekdiary.com/unix-linux-access-control-lists-acls-basics/? Don't know if it's good, it's just the first Google result 😛)

docs/Userguide_sharing_data.rst
```diff
@@ -0,0 +1,101 @@
...
+
+As an illustrative example, to allow ``$USER`` to share with ``$USER2`` in
+``rwx`` a hierarchy ``/network/scratch/$USER/X/Y/Z/...`` with ACLs:
```
vict0rsch, line 5: I'm not sure "in `rwx`" would be comprehensible to a beginner

docs/Userguide_sharing_data.rst
```diff
@@ -0,0 +1,101 @@
...
+
+----
+
+| Grant **oneself** permissions to access any **future** files/folders created
+  by the other *(or oneself)*
```
vict0rsch, line 10: The `$USER` / `$USER2` paradigm in the intro is good to understand what's up, this `oneself` / `other` / `another` version is less clear IMHO

vict0rsch: Also, I couldn't understand this sentence before the 4th read.

Is this saying: "how to make sure _I_ (`$USER`), will be able to `rwx` the files created by someone else (`$USER2`) if they put anything in `/network/scratch/$USER/X/Y/Z/`"

If that's the case, I think the first item in this tutorial should be how to _allow_ `$USER2` to even have access to `/network/scratch/$USER/X/Y/Z/`

docs/Userguide_sharing_data.rst
```diff
@@ -0,0 +1,101 @@
...
+| (``-d`` renders this permission a "default" / inheritable one)
+
+.. code-block:: bash
+
+   setfacl -Rdm user:$USER:rwx  /network/scratch/$USER/X/Y/Z/
```
vict0rsch, line 15: I'm guessing `-R` means recursive? `-d` is explained, what about `-m`?

docs/Userguide_sharing_data.rst
```diff
@@ -0,0 +1,101 @@
...
+  by the other *(or oneself)*
+| (``-d`` renders this permission a "default" / inheritable one)
+
+.. code-block:: bash
+
+   setfacl -Rdm user:$USER2:rwx /network/scratch/$USER/X/Y/Z/
+
+----
+
+| Grant **another** permission to access any **existing** files/folders created
```
vict0rsch, line 29: I would put how to handle *existing files* _before_ *future files*

docs/Userguide_sharing_data.rst
```diff
@@ -0,0 +1,101 @@
...
+  their creation.
+
+.. code-block:: bash
+
+   setfacl -Rm  user:$USER2:rwx /network/scratch/$USER/X/Y/Z/
+
+----
+
+| Grant **another** permission to search through one's hierarchy down to the
+  shared location in question.
```
vict0rsch, line 42: Why would we want to do that? What does "search" mean? Only `ls`? What about reading?

docs/Userguide_sharing_data.rst
```diff
@@ -0,0 +1,101 @@
...
+
+* **Non**-recursive (!!!!)
+* May also grant ``:rx`` in unlikely event others listing your folders on the
+  path is not troublesome or desirable.
+
+----
+
+| In order to access a file, all folders from the root (``/``) down to the
```
vict0rsch, line 50: Is this part of the previous section on "searching"? It _looks_ like it's its own different example, but the content seems like it's part of the previous block

vict0rsch:
Thank you for working on this! I've made a bunch of comments. I think this is still a little raw and rough for beginners to understand completely though. 

In particular, I think this PR needs an explicit section on **Projects** and the ideal workflow for collaborators not to get confused with permissions and people not having access to others' data etc. I'm thinking about a "Good practices for Projects" section with the elements mentioned in #115: 

* group creation request to it-support
* profs approve
* provide usernames
* **how to setup ACLs for user groups**

(On that note, can you not make it default that any and all members of a group have `rwx` on the project's folder `/network/projects/my_project` so that by default people don't have to do anything with ACLs?)

docs/Userguide_sharing_data.rst
```diff
@@ -0,0 +1,101 @@
...
+
+.. code-block:: bash
+
+   setfacl -m   user:$USER2:x   /network/scratch/$USER/X/Y/
+   setfacl -m   user:$USER2:x   /network/scratch/$USER/X/
+   setfacl -m   user:$USER2:x   /network/scratch/$USER/
+
+----
+
+| To bluntly allow **all** users to search through a folder (**think twice!**),
```
vict0rsch, line 65: What if I want them to be able to copy (read) a file? Or write?

vict0rsch: (_I_ personally know, I'm saying those examples should be there too 😄)

docs/Userguide_sharing_data.rst
obilaniu: @vict0rsch There is a careful reasoning I made for executing those commands in that specific order, it's not by accident.

The idea behind granting *yourself* access to all future files created by (you or other) *first* is that if you don't do that, then in the gap between your execution of that command and the next one, that other person could race you to create files you mysteriously can't delete.

Granting rights to even access the shared workspace to the other person *last* ensures that all ACLs have been readied beforehand and no race condition is possible at the moment of the "unlock".

docs/Userguide_sharing_data.rst
obilaniu: Yes, `-R` is recursive. `-m` means "modify" (add that ACL rule to the set), as opposed to "set" (replace all ACL rules with the ones provided).

More on that in `man setfacl`. I think people should develop at least some reflex to consult the man-pages, it's almost always `man commandname`.

docs/Userguide_sharing_data.rst
vict0rsch: So this is for security reasons? It's interesting. Put this^ as a `..note:` no? A good doc is also a good place to learn 😛 But I understand now thanks

docs/Userguide_sharing_data.rst
obilaniu: See a comment above

docs/Userguide_sharing_data.rst
obilaniu: To put it in extremely plain language, Unix separates the concept of "traversing" a directory from "reading" it.

To execute `ls /network/scratch/$USER/a/b/c/workspace/`, a `$USER2` **does not** require "read" (`r`) permission on _anything_ except `workspace/` itself. A `$USER2` **does** however require "search" permissions on **all** directories from the root down to `workspace/`. Unix overloads the "executable" (`x`) permission (which makes no sense on directories) for "search".

The rules are somewhat baroque but are elaborated upon in greater detail in `man path_resolution`. The summary that's important for most people is what I wrote above, and what my proposed `setfacl` commands arrange for.

btravouillon:
@obilaniu would you have some time available to update the PR in regards to @vict0rsch comments?

docs/Userguide_sharing_data.rst
```diff
@@ -0,0 +1,101 @@
...
+  This is already the case for all users for folders such as ``/``, ``/network``
+  and ``/network/scratch``, but users must explicitly grant access to some or
+  all users by adding ACLs for ``/network/scratch/$USER``, ``$HOME`` and
+  subfolders.
+
+.. code-block:: bash
+
+   setfacl -m   user:$USER2:x   /network/scratch/$USER/X/Y/
+   setfacl -m   user:$USER2:x   /network/scratch/$USER/X/
+   setfacl -m   user:$USER2:x   /network/scratch/$USER/
```
btravouillon, line 61: Just noticed the path should be `/network/scratch/$U{USER:0:1}/$USER/`

liellnima: Hey, this example here is very helpful - immediately solved my questions and it became quickly clear that one has to make the subdirectories accessible/searchable first!

btravouillon:
@obilaniu would you have some time available to update the PR in regards to @vict0rsch comments?

docs/Userguide_sharing_data.rst
obilaniu: We will make ourselves a task for this in #128.

docs/Userguide_sharing_data.rst
obilaniu: We will make ourselves a task for this in #128.

obilaniu (approved):

vict0rsch:
@saleml does not know what ACLs are. He'll be a great tester for this PR :)

docs/Userguide_sharing_data.rst
btravouillon: It's not for security reasons but for safety reasons. Not sure we should add too much explanations about the examples.

docs/Userguide_sharing_data.rst
btravouillon: People need to use the man-pages to get further explanation about commands and parameters. The documentation emphasizes the important parameters in the context, but we should not have to duplicate the man pages by explaining each parameter.

docs/Userguide_sharing_data.rst
obilaniu: @btravouillon In the end I did supply such an explanation, see [here](https://mila-docs-staging--121.org.readthedocs.build/en/121/Userguide.html#sharing-data-with-acls)

I think the current notes strike the right balance between explaining why this should probably not be skipped and delving too deep in the weeds.

btravouillon: Well, the comment was not resolved. I just noticed while doing the (hopefully) last review.

btravouillon (approved):
Looks good to me. 👍

PR #124, opened 2022-06-16
Update compute specs

btravouillon:
- HT/SMT was disabled on the `cn` nodes, there is a single thread per
   core now.
 - The DGX-1 `milaXX` were renamed `cn-eXXX`.
 - Old K80 and M40 were removed (`kepler[3-4]`)
 - rtx[2,6,8,9] are not available anymore


docs/Information_nodes.rst
```diff
@@ -21,37 +21,31 @@ Node profile description
...
 | :h:`<h5 style="margin: 5px 0 0 0;">Legacy GPU</h5>`                                                                                                             |
 +---------------------------------------+----------+---+------+---------+--------------+--------------+-------------+--------------+--------+---------------------+
-| kepler3                               | k80      | 8 |  16  |    2    |       4      |       2      |     256     |      3.6     | x86_64 |     kepler,12gb     |
-+---------------------------------------+----------+---+------+---------+--------------+--------------+-------------+--------------+--------+---------------------+
-| kepler4                               | m40      | 4 |  16  |    2    |       4      |       2      |     256     |      3.6     | x86_64 |    maxwell,24gb     |
-+---------------------------------------+----------+---+------+---------+--------------+--------------+-------------+--------------+--------+---------------------+
 | kepler5                               | V100     | 2 |  16  |    2    |       4      |       2      |     256     |      3.6     | x86_64 |      volta,16gb     |
```
satyaog, line 40: Labeling a v100 in a Legacy GPU section might bring confusion as to why other v100 are not also legacy? Maybe Legacy Board/Machine could be a better name now?

satyaog (approved):

docs/Information_nodes.rst
btravouillon: Note the hostname is `kepler` while there is no K80 available, only V100... This is a legacy system. Should I update the description for "Legacy GPU compute nodes"?

satyaog: Yes at least for me that sounds like a better description

PR #127, opened 2022-08-29
Clarification suggestion

postylem:
I misunderstood `path/to/code` to mean the path to the executable `code` on my local machine (like `$(which code)`).  Instead I see this is supposed to be a path on the cluster.  Edited to make unambiguous.

### DIFF
--- a/docs/VSCode.rst
+++ b/docs/VSCode.rst
@@ -25,7 +25,7 @@ run:
 
 .. code-block:: bash
 
-    mila code path/to/code
+    mila code path/on/cluster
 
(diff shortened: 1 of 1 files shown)


satyaog:
@breuleux can you check if you could clarify a bit here?

breuleux (approved):
Indeed, that seems clearer.

btravouillon (approved):
Looks good to me. 👍

satyaog (approved):

btravouillon:
Thank you @postylem!

PR #130, opened 2022-10-04
replace `$VAR` with `${VAR}` in ACL commands

vict0rsch:
Because ZSH reads `$VAR:r` as `${VAR:r}` and "swallows" `setfacl`'s `:r`


obilaniu (approved):
Super!

PR #138, opened 2022-11-10
Use python 3.8 instead of 3.9 in the examples

obilaniu:
Our Lmod module `python/3.9.5` is currently broken, avoid recommending it in the examples (ref: FD ticket# 10328)


btravouillon (approved):
LGTM.

satyaog (approved):

PR #139, opened 2022-11-11
Update advice with respect to CPU nodes and partitions

### LINKED ISSUE #137: Update the documentation with the new partitions for CPU nodes
Along with the new `cn-f[001-004]` nodes, introduce also a more sophisticated CPU partition scheme.

 - Introduce `-cpu` variants of the `unkillable`, `main` and `long` partitions (+ -grace variants).
 - Translate GPU `unkillable`, `main` and `long` partition names (+ - grace variants) to their `-cpu` equivalents dynamically if no GPU requested. This allows the same partition name to work in both situations.
 - Translate legacy `cpu_jobs*` partition names to the equivalent new names dynamically and retire the old names.
 - Prefer `cn-f` nodes for CPU jobs over any other nodes, but do allow running `long-cpu(-grace)` jobs on GPU nodes as a second-class, fallback option using `long-cpu(-grace)-eek` partitions (so-named because we "eek!" in disgust at having to use them).

obilaniu:
Also makes a few minor corrections elsewhere.

Closes #137.


docs/Information_nodes.rst
```diff
@@ -62,6 +66,10 @@ commands::
 
     --gres=gpu:a100:<number> --reservation=DGXA100
 
+.. note::
+   If you use these nodes, you are asked to please acknowledge NVIDIA's gift in
+   in any academic work that has benefited from it.
```
btravouillon, line 71: It's not true for `cn-d[001-002]`, only for `cn-d[003-004]`. Let's keep this for another PR once the team will have set up the program.

docs/Userguide_multigpu.rst
```diff
@@ -15,7 +15,7 @@ without any notice. To allow a different preemption mechanism, every partition
 have been duplicated (i.e. have the same characteristics as their counterparts)
 allowing a **120sec** grace period before killing your job *but don't requeue
 it automatically*: those partitions are referred by the suffix: ``-grace``
-(``main-grace, low-grace, cpu_jobs-grace``).
+(``main-grace, low-grace, main-cpu-grace, long-cpu-grace``).
```
btravouillon, line 18: Woops, `low-grace` does not exist. Should it be `long-grace`?

docs/Userguide_running_code.rst
```diff
@@ -350,33 +368,22 @@ Special GPU requirements
 Specific GPU *architecture* and *memory* can be easily requested through the
 ``--gres`` flag by using either
 
-* ``--gres=gpu:architecture:memory:number``
```
btravouillon, line 353: Did we remove that? How can one request an A100 40GB vs. an A100 80GB?

btravouillon (changes requested):
A few comments in-line, but that's fine overall. Thanks!

docs/Userguide_running_code.rst
obilaniu: @btravouillon SLURM removed it. You can't supply a 4-part spec like that, it's rejected by SLURM's pre-parser as an invalid GRES spec before `job_submit.lua` can even get its paws on it. So I removed support for it from there long ago, as it is unreachable.

btravouillon (approved):
Looks good to me. 👍

PR #140, opened 2022-11-19
Add datasets to home page

### LINKED ISSUE #94: Finding dataset on the home page of the docs
Is there a page describing all the datasets on the cluster? Is this page easily available from the homepage: https://docs.mila.quebec/?

I am thinking about a table with dataset name and the absolute path on the cluster where we can find the dataset.

satyaog:
resolves #94


btravouillon (approved):
Looks good to me. 👍

obilaniu (approved):

PR #157, opened 2023-02-24
Add Mila acknowledgement

satyaog:


docs/Acknowledgement_text.rst
```diff
@@ -0,0 +1,19 @@
...
+   are published on research that used resources provided by the organization.
+   Our ability to showcase the successful use of our resources builds an
+   important case for long-term and sustainable funding of Mila and our
+   operations. Examples include the use, either in whole or in part, of:
+
+   * Training Models
+   * Exploring Datasets
+   * Computing, visualization or storage hardware
+   * Software
+   * Collaboration facilities
```
gyom, line 13: Does this become too wide and lose some of its significance?

Do we want to make this more about the Mila cluster, or is it fine to have it be a general "yay Mila!" acknowledgement, in which case that's a bit implied by the fact that people are using @mila.quebec in their email addresses?

docs/Acknowledgement_text.rst
```diff
@@ -0,0 +1,19 @@
...
+   * Staff expertise / consultation
```
satyaog, line 14:
```suggestion
   * Computing, visualization or storage hardware
   * Software from `<https://github.com/mila-iqia>`_
   * Staff expertise / consultation
```

docs/Acknowledgement_text.rst
satyaog: Yes, removed some of the bullets

satyaog:
quick urls :
* https://mila-docs--157.org.readthedocs.build/en/157/#mila-technical-documentation
* https://mila-docs--157.org.readthedocs.build/en/157/Acknowledgement.html
* https://mila-docs--157.org.readthedocs.build/en/157/Extra_compute.html#digital-research-alliance-of-canada-clusters

docs/Acknowledgement_text.rst
```diff
@@ -0,0 +1,16 @@
...
+   operations. Examples include the use, either in whole or in part, of:
+
+   * Computing, visualization or storage hardware
+   * Software from `<https://github.com/mila-iqia>`_
+   * Staff expertise / consultation
+
+   The exact wording of the acknowledgment may vary, but please ensure that
+   Mila is mentioned. The following can be used as a guideline:
+
+      *"This research was enabled in part by support provided by Mila (mila.quebec)."*
```
bouthilx, line 16: Do we want this acknowledgment for any kind of resources or computational resources specifically?

satyaog: I think it makes sense to have both? This is largely using the acknowledgement of CC which made sense to me as well: https://alliancecan.ca/en/services/advanced-research-computing/acknowledging-alliance

bouthilx: I think in the case of Compute Canada it's quite clear that the resources they provide is compute. Mila on the other hand provides lots of different resources to researchers. If what we mostly need to know if what papers used compute provided by Mila to support extending the cluster then I'd say the acknowledgement for Mila resources needs to be more precise than the one for Compute Canad.

satyaog: I'm not entirely sure why we wouldn't want to receive acknowledgement for software / consultation / general help as well (even if we might not have that much right now) but I've cleaned up the text

bouthilx: Indeed, it would also be useful to receive acknowledgement for these services. Maybe the sentence could include all these explicitly, and we could ask the researchers to adapt the sentence to whatever services they benefited from for their research.

btravouillon (approved):
Looks good to me. 👍

obilaniu (approved):

PR #160, opened 2023-03-07
Add "Minimal Examples" section with a PyTorch Setup example

lebrice:
WIP: Adding a new section with minimal examples.

Each example will contain the following:

- A list of **prerequisites**: Sections of the documentation that have to be read before using the example
- An sbatch script (`job.sh`)
- A Python script (`main.py`)
- A `README.rst` file, refered to in the documentation.

TODOs:
- [x] Remove the duplicated content from #151.
- [x] Add a "Running this example" section for each example.
- [x] Fix the "multiple labels X" (where X="prerequisites" or "Running this example") issue with the .rst files,

PR #161, opened 2023-03-09
Add Pytorch distributed training examples

### LINKED ISSUE #99: Examples on how to use `/network/datasets/cifar10` with pytorch dataset

lebrice:
Should be merged **after** #160 

Fixes #99


docs/examples/distributed/001_single_gpu/main.py
```diff
@@ -0,0 +1,172 @@
...
+        shuffle=False,
+    )
+
+    # Checkout the "checkpointing and preemption" example for more info!
+    logger.debug("Starting training from scratch.")
+
+    for epoch in range(training_epochs):
+        logger.debug(f"Starting epoch {epoch}/{training_epochs}")
+
+        # Set the model in training mode (this is important for e.g. BatchNorm and Dropout layers)
```
satyaog, line 69: To remove one diff

```suggestion
        # Set the model in training mode (important for e.g. BatchNorm and Dropout layers)
```

docs/examples/distributed/002_multi_gpu/main.py
```diff
@@ -0,0 +1,271 @@
...
+    print("Done!")
+
+
+@torch.no_grad()
+def validation_loop(model: nn.Module, dataloader: DataLoader, device: torch.device):
+    model.eval()
+
+    total_loss = torch.as_tensor(0.0, device=device)
+    n_samples = torch.as_tensor(0, device=device)
+    correct_predictions = torch.as_tensor(0, device=device)
```
satyaog, line 168: This is for my personal comprehension but why are those lines different from the base scrpit?

docs/examples/distributed/002_multi_gpu/main.py
```diff
@@ -0,0 +1,271 @@
...
+        batch = tuple(item.to(device) for item in batch)
+        x, y = batch
+
+        logits: Tensor = model(x)
+        loss = F.cross_entropy(logits, y)
+
+        batch_n_samples = x.shape[0]
+        batch_correct_predictions = logits.argmax(-1).eq(y).sum()
+
+        total_loss += loss
```
satyaog, line 180: This is for my personal comprehension but why is this line different from the base scrpit?

docs/examples/distributed/003_multi_node/README.rst
```diff
@@ -0,0 +1,40 @@
...
+* :ref:`pytorch_setup`
+* :ref:`001 - Single GPU Job`
+* :ref:`002 - Multi-GPU Job`
+
+Other interesting resources:
+* `https://sebarnold.net/dist_blog/`
+* `https://lambdalabs.com/blog/multi-node-pytorch-distributed-training-guide`
+
+
+Click here to see `the source code for this example <https://github.com/lebrice/mila-docs/tree/pytorch_distributed_training_examples/docs/examples/distributed/003_multi_node>`_
```
satyaog, line 16: Should probably use `mila-iqia`
```suggestion
Click here to see `the source code for this example <https://github.com/mila-iqia/mila-docs/tree/pytorch_distributed_training_examples/docs/examples/distributed/003_multi_node>`_
```

docs/examples/distributed/001_single_gpu/README.rst
```diff
@@ -0,0 +1,29 @@
+001 - Single GPU Job
+====================
+
+
+**Prerequisites**
+Make sure to read the following sections of the documentation before using this example:
+
+* :ref:`pytorch_setup`
+
+The full source code for this example is available on `the mila-docs GitHub repository. <https://github.com/lebrice/mila-docs/tree/pytorch_distributed_training_examples/docs/examples/distributed/001_single_gpu>`_
```
satyaog, line 10: Should probably use `mila-iqia`
```suggestion
The full source code for this example is available on `the mila-docs GitHub repository. <https://github.com/mila-iqia/mila-docs/tree/pytorch_distributed_training_examples/docs/examples/distributed/001_single_gpu>`_
```

docs/examples/distributed/003_multi_node/job.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+#SBATCH --gpus-per-task=rtx8000:1
+#SBATCH --cpus-per-task=1
```
satyaog, line 3: Shouldn't this this be like in 002?
```suggestion
#SBATCH --cpus-per-task=4
```

docs/examples/distributed/003_multi_node/main.py
```diff
@@ -0,0 +1,279 @@
...
+    weight_decay = 1e-4
+    batch_size = 128  # NOTE: This is the "local" batch size, per-GPU.
+
+    # Check that the GPU is available
+    assert torch.cuda.is_available() and torch.cuda.device_count() > 0
+    rank, world_size, local_rank = setup()
+    is_master = rank == 0
+    is_local_master = local_rank == 0
+    device = torch.device("cuda", local_rank)
+    import rich.logging
```
satyaog, line 29: Was there some issues with having `import rich.logging` at the top like in 002?

satyaog (commented):
All this is very cool and interesting!

docs/examples/distributed/001_single_gpu/main.py
```diff
@@ -0,0 +1,172 @@
...
+
+            # Advance the progress bar one step, and update the "postfix" () the progress bar. (nicer than just)
+            progress_bar.update(1)
+            progress_bar.set_postfix(loss=loss.item(), accuracy=accuracy.item())
+        progress_bar.close()
+
+        val_loss, val_accuracy = validation_loop(model, valid_dataloader, device)
+        logger.info(f"Epoch {epoch}: Val loss: {val_loss:.3f} accuracy: {val_accuracy:.2%}")
+
+    print("Done!")
```
satyaog, line 109: Should there be a line to copy the trained model somewhere?

docs/examples/distributed/001_single_gpu/main.py
```diff
@@ -0,0 +1,172 @@
...
+            optimizer.zero_grad()
+            loss.backward()
+            optimizer.step()
+
+            # Calculate some metrics:
+            n_correct_predictions = logits.detach().argmax(-1).eq(y).sum()
+            n_samples = y.shape[0]
+            accuracy = n_correct_predictions / n_samples
+
+            logger.debug(f"Accuracy: {accuracy.item():.2%}")
```
Delaunay, line 98: When logging values people should avoid using interpolation and use ` logger.debug("Accuracy: %.2f", accuracy.item())` instead because the later is actually lazy so the string building does not happen unless the log level is sufficient

obilaniu: `.item()` causes the bulk of the slowdown already and implies CUDA synchronization if it's a CUDA tensor. Converting it to a string in this thread afterwards will cost nanoseconds and generate at most 20ish new characters (and here, probably about 4).

I consider laziness here an anti-feature. If it's true that it's lazy, I'm horrified at the possibility of somebody passing a CUDA scalar accidentally to another process, especially one that has `fork()`'ed and is in a bad-CUDA-fork state.

Is the performance difference between `wandb` stitching internally 100 `"Accuracy: "` + 100 `"%.2f"` and `"Accuracy: %.2f"` strings together noteworthy?

Delaunay: That might not matter in that case but it is a good habit to take

docs/examples/distributed/002_multi_gpu/main.py
```diff
@@ -0,0 +1,271 @@
...
+            # Calculate some metrics:
+            # local metrics
+            local_n_correct_predictions = logits.detach().argmax(-1).eq(y).sum()
+            local_n_samples = logits.shape[0]
+            local_accuracy = local_n_correct_predictions / local_n_samples
+
+            # "global" metrics: calculated with the results from all workers
+            # NOTE: Creating new tensors to hold the "global" values, but this isn't required.
+            n_correct_predictions = local_n_correct_predictions.clone()
+            torch.distributed.all_reduce(n_correct_predictions, op=torch.distributed.ReduceOp.SUM)
```
Delaunay, line 131: Probably not that impactful but
Because only the master needs the final value for logging purposes you could switch `all_reduce` to `reduce`
that might avoid some unnecessary data transfer

docs/examples/distributed/001_single_gpu/main.py
```diff
@@ -0,0 +1,172 @@
...
+from torchvision.datasets import CIFAR10
+from torchvision.models import resnet18
+from tqdm import tqdm
+
+
+def main():
+    training_epochs = 10
+    learning_rate = 5e-4
+    weight_decay = 1e-4
+    batch_size = 128
```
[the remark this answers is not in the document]
lebrice: > Should we add some lines regarding setting the random seed?

Hmmm I was thinking that we could add an example for reproduciblity, in which case that example would have all things required for 100% reproducible training (here just setting the seed would not be sufficient, in paritcular with distributed training I think)

docs/examples/distributed/001_single_gpu/main.py
```diff
@@ -0,0 +1,172 @@
...
+            n_correct_predictions = logits.detach().argmax(-1).eq(y).sum()
+            n_samples = y.shape[0]
+            accuracy = n_correct_predictions / n_samples
+
+            logger.debug(f"Accuracy: {accuracy.item():.2%}")
+            logger.debug(f"Average Loss: {loss.item()}")
+
+            # Advance the progress bar one step, and update the "postfix" () the progress bar. (nicer than just)
+            progress_bar.update(1)
+            progress_bar.set_postfix(loss=loss.item(), accuracy=accuracy.item())
```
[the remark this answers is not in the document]
lebrice: I'd like to avoid adding additional dependencies.

docs/examples/distributed/002_multi_gpu/main.py
lebrice: They are tensors since we'll do a reduce across workers.

docs/examples/test_examples.py
```diff
@@ -0,0 +1,61 @@
...
+    conda_dir = tmp_path_factory.mktemp("conda_envs")
+    conda_env_prefix = conda_dir / "pytorch"
+
+    # TODO: Modify the file so we create the conda environment in $SLURM_TMPDIR.
+    with open("docs/examples/Frameworks/pytorch_setup/job.sh") as f:
+        job = f.read()
+
+    assert "# conda create -y -n pytorch" in job
+    assert "#     pytorch-cuda=11.6" in job
+    assert "conda activate pytorch" in job
```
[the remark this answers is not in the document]
lebrice: Removed this file in 77e64de41d528266191c4b579fc01e08d6e37d8a

docs/examples/test_examples.py
```diff
@@ -0,0 +1,61 @@
...
+    job = job.replace("conda activate pytorch", f"conda activate {conda_env_prefix}")
+
+    sbatch_script_path = Path(conda_dir / "setup_job.sh")
+    sbatch_script_path.write_text(job)
+
+    # TODO: Actually run the job that creates the conda environment.
+    job_id = subprocess.check_output(
+        ["sbatch", "--parsable", f"--output={conda_dir}/setup_job.out", str(sbatch_script_path)]
+    )
```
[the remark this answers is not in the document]
lebrice: Removed this file in 77e64de41d528266191c4b579fc01e08d6e37d8a

docs/examples/test_examples.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_pytorch_setup(monkeypatch: pytest.MonkeyPatch, pytorch_conda_env: Path, tmp_path: Path):
```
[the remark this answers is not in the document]
lebrice: Removed this file in 77e64de41d528266191c4b579fc01e08d6e37d8a

docs/examples/frameworks/pytorch_setup/README.rst
```diff
@@ -0,0 +1,35 @@
...
+===================
+
+.. IDEA: Add a link to all the sections of the documentation that have to
+.. absolutely have been read before this tutorial.
+
+**Prerequisites**: (Make sure to read the following before using this example!)
+
+* :ref:`Quick Start`
+* :ref:`Running your code`
+* :ref:`Conda`
```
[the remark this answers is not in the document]
lebrice: The comment explains the idea behind the Prerequisites section. I can remove it, sure.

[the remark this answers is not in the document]
lebrice: Fixed in f5592ee3b020398c527f608699bf3eec0b57b52b

docs/examples/distributed/003_multi_node/main.py
lebrice: Fixed in c23e81e31de1618db4aa0930608fd3aa6e9b3e8f

docs/examples/distributed/001_single_gpu/main.py
lebrice: That should probably be part of the checkpointing tutorial, no?

docs/examples/distributed/002_multi_gpu/main.py
lebrice: Fixed in 7c0a5f5b76ab65c49f7333f8222d51018d89e7a8

docs/examples/generate_diffs.sh
```diff
@@ -0,0 +1,36 @@
...
+    # TODO: Create a diff file that can then be used to make writing new examples easier.
+    git diff --no-index \
+        "examples/$1" \
+        "examples/$2" \
+        | grep -Ev "^index" \
+        > examples/$2.diff
+
+    # Write a diff file to be shown in the documentation.
+    
+    built_example_diffs_folder="_build/example_diffs"
```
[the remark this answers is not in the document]
lebrice: I'm not sure I understand. What value should I put here?
The idea is just for this to save the diffs somewhere, and then the `README.rst`s reference them.

docs/examples/distributed/001_single_gpu/main.py
```diff
@@ -0,0 +1,172 @@
...
+    # Create a model and move it to the GPU.
+    model = resnet18(num_classes=10)
+    model.to(device=device)
+
+    optimizer = torch.optim.AdamW(model.parameters(), lr=learning_rate, weight_decay=weight_decay)
+
+    # Setup CIFAR10
+    num_workers = get_num_workers()
+    dataset_path = os.environ.get("SLURM_TMPDIR", "../dataset")
+    train_dataset, valid_dataset, test_dataset = make_datasets(dataset_path)
```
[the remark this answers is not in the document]
lebrice: Fixed in c23e81e31de1618db4aa0930608fd3aa6e9b3e8f

docs/examples/distributed/001_single_gpu/job.sh
```diff
@@ -0,0 +1,35 @@
...
+#     pytorch-cuda=11.6 -c pytorch -c nvidia
+# Other conda packages:
+# conda install -y -n pytorch -c conda-forge rich
+
+# Activate pre-existing environment.
+conda activate pytorch
+
+
+# Stage dataset into $SLURM_TMPDIR
+cp -a /network/datasets/cifar10.var/cifar10_torchvision $SLURM_TMPDIR
```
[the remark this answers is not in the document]
lebrice: Fixed in c23e81e31de1618db4aa0930608fd3aa6e9b3e8f

docs/examples/distributed/001_single_gpu/main.py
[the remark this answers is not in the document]
lebrice: Actually fixed in 42273315aef259c0dc83a4af24e1c57b66ae6f22

docs/examples/distributed/002_multi_gpu/main.py
lebrice: Actually fixed in fc0a8d910520dd9a911c1d6c94539b080200c138

docs/examples/generate_diffs.sh
```diff
@@ -0,0 +1,36 @@
...
+    echo "Generating diff for docs/examples/$1 -> docs/examples/$2"
+    # NOTE: Assuming that this gets run from the `docs` folder (as is the case when building the docs).
+
+    # TODO: Create a diff file that can then be used to make writing new examples easier.
+    git diff --no-index \
+        "examples/$1" \
+        "examples/$2" \
+        | grep -Ev "^index" \
+        > examples/$2.diff
+
```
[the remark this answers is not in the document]
lebrice: Removed this in 90656658160099396bfa3d165590f74e7d427ffe

docs/examples/generate_examples.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+function generate_example() {
+    patch  docs/examples/$1 -i docs/examples/$2.diff -o docs/examples/$2
```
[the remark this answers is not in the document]
lebrice: Removed this script in 58d22ad27e95293204e49e8834751722978b100e

docs/examples/distributed/002_multi_gpu/job.sh
```diff
@@ -0,0 +1,43 @@
...
+# Other conda packages:
+# conda install -y -n pytorch -c conda-forge rich tqdm
+
+# Activate pre-existing environment.
+conda activate pytorch
+
+
+# Stage dataset into $SLURM_TMPDIR
+mkdir -p $SLURM_TMPDIR/data
+ln -s /network/datasets/cifar10/cifar-10-python.tar.gz $SLURM_TMPDIR/data/
```
obilaniu, line 33: `ln -s` fait un symbolic link. Ceci redirigera tous les IOs sur `/network/datasets`. À ne pas recommender en général....

docs/examples/distributed/003_multi_node/job.sh
```diff
@@ -0,0 +1,43 @@
...
+#     pytorch-cuda=11.7 -c pytorch -c nvidia
+# Other conda packages:
+# conda install -y -n pytorch -c conda-forge rich tqdm
+
+# Activate pre-existing environment.
+conda activate pytorch
+
+# Stage dataset into $SLURM_TMPDIR (only on the first worker of each node)
+srun --ntasks=$SLURM_JOB_NUM_NODES --ntasks-per-node=1 bash -c \
+   'mkdir -p $SLURM_TMPDIR/data && ln -s /network/datasets/cifar10/cifar-10-python.tar.gz $SLURM_TMPDIR/data/'
```
obilaniu, line 33: Ça c'est bien par contre.

docs/examples/distributed/003_multi_node/main.py
```diff
@@ -0,0 +1,291 @@
...
+    # - others: Barrier THEN *NO* Download
+    if not is_master:
+        # Wait for the master process to finish downloading (reach the barrier below)
+        torch.distributed.barrier()
+    train_dataset = CIFAR10(
+        root=dataset_path, transform=transforms.ToTensor(), download=is_master, train=True
+    )
+    test_dataset = CIFAR10(
+        root=dataset_path, transform=transforms.ToTensor(), download=is_master, train=False
+    )
```
obilaniu, line 267: Vu que nous avons utilisé `srun` pour faire le staging, y a-t-il lieu de faire `download=False` point final?

lebrice: Malheureusement il faut passer `download=True` pour que torchvision extraie l'archive si elle est déjà là.

docs/examples/distributed/002_multi_gpu/job.sh
lebrice: Je le fais juste pour l'archive par contre, donc c'est correct ici?

docs/examples/distributed/003_multi_node/main.py
obilaniu: Oh hmm ok

docs/examples/distributed/002_multi_gpu/job.sh
obilaniu: _Okay_, ici et dans ce contexte, oui c'est techniquement la meilleure chose à faire, mais ça mène en erreur parce que ça ne s'applique qu'au contexte des torchvision DataLoaders. Si quelqu'un fait copié-collé de ce template pour son test avec un méga dataset alternatif, et que le data loader sait comment lire "direct" le Zip/Tar whatever sans unpack (comme [ici](https://github.com/MLRichter/ConvNeXt/blob/benchmark/nested_tarbal_parser.py)), les I/Os iront directement sur `$SCRATCH`, sans que l'utilisateur plus novice n'en devienne conscient.

Je préférerais `cp` ici pour symboliser "copie" ici au sens direct, mais peut-être avec un commentaire qui inclut une commande `tar` et `zip`:

```bash
# Stage dataset into $SLURM_TMPDIR
mkdir -p $SLURM_TMPDIR/data
cp /network/datasets/cifar10/cifar-10-python.tar.gz $SLURM_TMPDIR/data/
# General-purpose alternatives combining copy and unpack:
#     unzip   /network/datasets/some/file.zip -d $SLURM_TMPDIR/data/
#     tar -xf /network/datasets/some/file.tar -C $SLURM_TMPDIR/data/
```

docs/examples/distributed/001_single_gpu/main.py
satyaog: Yes it should be there in checkpointing to copy the model before the end of the training but should it also be present in those examples? Otherwise what do you do without the model?

docs/examples/distributed/002_multi_gpu/job.sh
lebrice: Fixed in 0e6c98951e10e717b25f3770f0fad0382e3a6797

obilaniu (approved):

docs/examples/generate_diffs.sh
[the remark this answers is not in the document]
satyaog: Actually, the only thing is that you put the files in a directory that could be something else. In itself, adding files in `_build` is not wrong but you could end up with 2 out dirs if something else is used to hold to build of the doc (e.g. you could end up with a `_build` dir containing only the `.diff` files as well as an `OUT_DIR` containing the build).

It's just not completely clean as something like `example_diffs` could be more explicit or adding the `.diff` directly in the example directory with some `.gitignore` files could be a solution as well. We can see later if we wish to change that

docs/examples/generate_examples.sh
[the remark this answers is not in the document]
satyaog: Would it have worked actually without getting conflicts? We might need to do a `git add -p` anyway but yes this script would have helped having a base for that I suppose

docs/examples/distributed/003_multi_node/main.py
```diff
@@ -0,0 +1,291 @@
...
+            loss.div_(world_size)  # Report the average loss across all workers.
+
+            accuracy = n_correct_predictions / n_samples
+
+            logger.debug(f"(local) Accuracy: {local_accuracy:.2%}")
+            logger.debug(f"(local) Loss: {local_loss.item()}")
+            # NOTE: This would log the same values in all workers. Only logging on master:
+            if is_master:
+                logger.debug(f"Accuracy: {accuracy.item():.2%}")
+                logger.debug(f"Average Loss: {loss.item()}")
```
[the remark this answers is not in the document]
lebrice: I agree, I think it's fine here. Perhaps if we have an example about performance or similar, we could start to care about this, but I currently don't really mind it.

docs/examples/distributed/001_single_gpu/main.py
lebrice: This is a multi-gpu or multi-node training script that doesn't handle pre-emption or checkpointing.

I think it should be relatively easy for users to combine the checkpointing example "diff" with the multi-gpu example "diff" to get a pre-emption-aware, multi-gpu training script.

satyaog: That's a detail in my mind, but I don't think we would be handling pre-emption or checkpointing if we only copy the model at the very end of the training. Let's go with want we have!

satyaog (approved):

lebrice:
> This one I'm a bit more uneasy as if we don't show to efficiently train a model using the GPU in a multi-gpu / multi-node example, where should we put it?

Hmmm I don't quite understand. What makes you think that this "doesn't show how to efficiently train a model using the GPU" ?
The impact on performance of doing this print (with one extra cpu-gpu sync) seems pretty minimal, and this sort of code is exactly what people are doing in practice. If we want to have a "super extra performance-y example", we can add one too, but this seems pretty reasonable to me at least.

satyaog:
Are people printing for each batch? Because that seams to be the main point here. We should do some tests to validate but the fact that PT to put this in the top it's GPU specific optimizations seams to indicate that this should be a frequent cause of slow down. Plus, it doesn't seam to be much work to reduce the number of syncs right?

obilaniu:
I don't _like_ it, but there are few to no clean ways of printing _something_ off the GPU without synchronizing. It's the nature of the beast: CUDA kernels are asynchronous, and if you want to write a demo to see that things work, it's either somewhat suboptimal, or it's complex enough that it's not a demo.

For the purpose of _this_ tutorial for newbies to distributed DL, I'd prefer to avoid complexity _unless_ it's directly related to the learning topic - the distributed aspect. I don't like encouraging bad practices, but a tutorial that is too messy might end up entirely ignored by the students.

PR #163, opened 2023-03-14
Big Dataset Examples

satyaog:

PR #164, opened 2023-03-14
Checkpointing Example

satyaog:


satyaog:
quick url : https://mila-docs--164.org.readthedocs.build/en/164/Minimal_examples.html#checkpointing

btravouillon:
@satyaog Is it expected to see an overlap with #163 in the changes? A lot of files include the same changes in both PR.

btravouillon:
Waiting for merge of #161.

satyaog:
There a lot of overlapping with #163. This one should be rebased on #163 once done

.github/workflows/tests.yml
```diff
@@ -0,0 +1,22 @@
...
+    - uses: actions/checkout@v2
+
+    - name: Set up Python
+      uses: actions/setup-python@v1
+
+    - name: Install dependencies
+      run:  python3 -m pip install -r docs/requirements.txt
+
+    - name: Run files generation tests
+      run:  pre-commit run --all-files && [[ -z "$(git status -s)" ]]
```
lebrice, line 22: What's  `&& [[ -z "$(git status -s)" ]]` for?

docs/examples/preprocess.py
```diff
@@ -0,0 +1,55 @@
...
+GitHub doesn't support include of other files, even of the same type and
+location, so this file generates a README.rst with files content embedded
+"""
+from glob import glob
+from pathlib import Path
+import shutil
+import subprocess
+
+
+def preprocess():
```
lebrice, line 11: Wow this is really cool!

One question though. If this gets called to generate the README.rst files automatically, then why are they included in the PR? Shouldn't this only include the _index.rst files?

docs/examples/frameworks/pytorch_setup/_index.rst
lebrice, line 1: Nitpick: Shouldn't this show up as a rename from README.rst to _index.rst rather than as creating a new file?
Did any of the content change?

.pre-commit-config.yaml
lebrice, line 1: Why use pre-commit to run the python script? Isn't it sufficient that the script is called in the spinx configuration file?

docs/examples/data/torchvision/job.sh
```diff
@@ -0,0 +1,54 @@
...
+# faster training
+srun --ntasks=$SLURM_JOB_NUM_NODES --ntasks-per-node=1 \
+	time -p bash data.sh "/network/datasets/inat" "$SLURM_TMPDIR/data" ${_DATA_PREP_WORKERS}
+
+
+# Fixes issues with MIG-ed GPUs with versions of PyTorch < 2.0
+unset CUDA_VISIBLE_DEVICES
+
+# Execute Python script
+python main.py
```
lebrice, line 54: If you have a multi-node dataset preparation routine with `srun`, then you'd probably also run the `python main.py` script on each node with `srun` (if not once per GPU)

docs/examples/data/checkpointing/job.sh
```diff
@@ -0,0 +1,59 @@
...
+# General-purpose alternatives combining copy and unpack:
+#     unzip   /network/datasets/some/file.zip -d $SLURM_TMPDIR/data/
+#     tar -xf /network/datasets/some/file.tar -C $SLURM_TMPDIR/data/
+
+
+# Fixes issues with MIG-ed GPUs with versions of PyTorch < 2.0
+unset CUDA_VISIBLE_DEVICES
+
+# Execute Python script
+python main.py
```
lebrice, line 59:
```suggestion
srun python main.py
```

docs/examples/data/checkpointing/main.py
```diff
@@ -0,0 +1,221 @@
...
+from torchvision import transforms
+from torchvision.datasets import CIFAR10
+from torchvision.models import resnet18
+from tqdm import tqdm
+
+
+try:
+    _CHECKPTS_DIR = f"{os.environ['SCRATCH']}/checkpoints"
+except KeyError:
+    _CHECKPTS_DIR = "../checkpoints"
```
lebrice, line 20:
```suggestion
_CHECKPOINTS_DIR = os.environ.get("SCRATCH", "..") + "/checkpoints"
```

docs/examples/data/checkpointing/main.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def main():
+    training_epochs = 10
+    learning_rate = 5e-4
+    weight_decay = 1e-4
+    batch_size = 128
+    resume_file = f"{_CHECKPTS_DIR}/resnet18_cifar10/checkpoint.pth.tar"
```
lebrice, line 28:
```suggestion
    resume_file = f"{_CHECKPOINTS_DIR}/resnet18_cifar10/checkpoint.pth.tar"
```

docs/examples/data/checkpointing/main.py
```diff
@@ -0,0 +1,221 @@
...
+        logger.debug(f"=> loading checkpoint '{resume_file}'")
+        # Map model to be loaded to gpu.
+        checkpoint = torch.load(resume_file, map_location="cuda:0")
+        start_epoch = checkpoint["epoch"]
+        best_acc = checkpoint["best_acc"]
+        # best_acc may be from a checkpoint from a different GPU
+        best_acc = best_acc.to(device)
+        model.load_state_dict(checkpoint["state_dict"])
+        optimizer.load_state_dict(checkpoint["optimizer"])
+        logger.debug(f"=> loaded checkpoint '{resume_file}' (epoch {checkpoint['epoch']})")
```
lebrice, line 58: This should also restore the torch/numpy/random RNG states from the saved checkpoint if possible.

docs/examples/data/checkpointing/main.py
```diff
@@ -0,0 +1,221 @@
...
+        is_best = val_accuracy > best_acc
+        best_acc = max(val_accuracy, best_acc)
+
+        save_checkpoint({
+            "epoch": epoch + 1,
+            "arch": "resnet18",
+            "state_dict": model.state_dict(),
+            "best_acc": best_acc,
+            "optimizer": optimizer.state_dict(),
+        }, is_best)
```
lebrice, line 149: Should also include the torch/numpy/random RNG states if possible.

lebrice (changes requested):
Not completely done reviewing, moving to #163  before I complete this.

.github/workflows/tests.yml
satyaog: making sure that nothing changed :

* `&&` : run the next command only if the previous succeeded
* `git status -s` : only print the files that differs (including new and deleted)
* `[[ -z ... ]]` : checks if the length of the output or variable is zero and returns true if it is zero.

docs/examples/preprocess.py
satyaog: Only because otherwise the README would be visible on github

docs/examples/frameworks/pytorch_setup/_index.rst
satyaog: It should but I think git prioritizes the name of the file rather than the content. I'm not sure how to force another behaviour but yes, that is annoying. I could try with `git mv` and see if the listing changes

satyaog:
Rebased on #182 which should be merged before this one

.github/workflows/tests.yml
lebrice: Note to self, we should do `pre-commit run --all-files preprocess-py` to only run this hook in particular.
(I locally added other hooks in my pre-commit config. If we ever add some other hooks for writing the examples (e.g. ruff, black, etc, then we should probably only run the generate hook here)

lebrice (approved):
(NOTE: J'approve ce qui est devenu ma propre PR, il me faudrait 1-2 autres reviews svp)

docs/examples/checkpointing/job.sh
```diff
@@ -0,0 +1,61 @@
...
+                            # https://dhruveshp.com/blog/2021/signal-propagation-on-slurm/
+
+# trap the signal to the main BATCH script here.
+sig_handler()
+{
+	echo "BATCH interrupted"
+	wait # wait for all children, this is important!
+}
+
+trap 'sig_handler' SIGINT SIGTERM SIGCONT
```
obilaniu: Useless after `exec python main.py`

lebrice: Yeah you're right! Thanks @obilaniu. Should we be encouraging users to do `srun python (...)` or `exec python (...)` in your view?

obilaniu: This being an example only, I would adopt one of two strategies - either minimize the delta from the foundation mini-example to this one, or minimize the complexity of this one in the absolute.

I would prefer that you drop the `trap` business, and use the Python equivalent (the `signal` package).

However, I have no faith in asynchronous checkpointing at all and recommend only synchronous checkpointing. I would use the `signal` package only to demonstrate in the mini-example that Python can install signal handlers and execute Python code as a result.

lebrice: Ok, here's my take: **779413b8fc32907607f268c55391dc839965a75a**

btravouillon (approved):
I'm not able to review the python code, but that changes looks good to me overall.

obilaniu:
Overall this is fine, however I do not like the strategy for checkpointing. It's not safe.

Your implementation:
1. For writing:
    1. *If* it exists, move aside the current checkpoint to a `<PATH>.backup` file
    2. Writes a checkpoint to `<PATH>`
    3. Unlinks `<PATH>.backup`.
2. For reading:
    1. If `<PATH>.backup` exists, load that
    2. If `<PATH>` exists, load that
    3. Otherwise, start from scratch

You vulnerability is that if you are killed in step 2 during the *first* checkpoint's write, the reload will detect no `<PATH>.backup` but does detect the existence of the (_unbeknownst to it, corrupt_) checkpoint `<PATH>`, and fail to reload.

---

My preferred implementation:
1. For writing:
    1. Writes a checkpoint to `<PATH>.<tmpXXXXXXXX>`. The name should be unique to avoid collisions.
    2. Atomically move `<PATH>.<tmpXXXXXXXX>` into `<PATH>` using [`os.replace()`](https://docs.python.org/3/library/os.html#os.replace).
2. For reading:
    1. If `<PATH>` exists, load that
    2. Otherwise, start from scratch.

lebrice:
Hey @obilaniu , I implemented your suggested changes in 4d3fd15294f7ee91e6606967468e75ed5158ea78

obilaniu (approved):

PR #176, opened 2023-04-13
Update the documentation about DGX A100

btravouillon:
The reservation DGXA100 does not exist anymore. One should use constraints to run a job on a DGX.

### DIFF
--- a/docs/Information_nodes.rst
+++ b/docs/Information_nodes.rst
@@ -58,13 +58,21 @@ DGX A100
 .. _dgx_a100_nodes:
 
 DGX A100 nodes are NVIDIA appliances with 8 NVIDIA A100 Tensor Core GPUs. Each
-GPU has 40 GB of memory, for a total of 320 GB per appliance. The GPUs are
-interconnected via 6 NVSwitches which allows 4.8 TB/s bi-directional bandwidth.
+GPU has either 40 GB or 80 GB of memory, for a total of 320 GB or 640 GB per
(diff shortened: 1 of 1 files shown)


docs/Information_nodes.rst
```diff
@@ -58,13 +58,20 @@ DGX A100
 .. _dgx_a100_nodes:
 
 DGX A100 nodes are NVIDIA appliances with 8 NVIDIA A100 Tensor Core GPUs. Each
-GPU has 40 GB of memory, for a total of 320 GB per appliance. The GPUs are
-interconnected via 6 NVSwitches which allows 4.8 TB/s bi-directional bandwidth.
+GPU has either 40GB or 80GB of memory, for a total of 320GB or 640GB per
+appliance. The GPUs are interconnected via 6 NVSwitches which allows 4.8 TB/s
+bi-directional bandwidth. See the table above for the specifications of each
```
obilaniu, line 63: > ... which allow for 600 GB/s point-to-point bandwidth _(unidirectional)_ and a full bisection bandwidth of 4.8 TB/s _(bidirectional)_.

I don't like how NVIDIA selects the largest possible B/W number, which looks impressive but is misleading as to a single GPU's actual performance. For the regular user, knowing

1) The bandwidth of one GPU
2) That this bandwidth is available in both directions simultaneously (full-duplex)
3) That in a DGX-A100 there is full bisection bandwidth (all four pairs can communicate simultaneously at their full rate)

is probably enough.

obilaniu (approved):

PR #188, opened 2023-05-05
Add weights doc

satyaog:


docs/Information_storage.rst
```diff
@@ -131,3 +132,19 @@ following command:
 .. prompt:: bash $
 
    ssh [CLUSTER_LOGIN] -C "projects/rrg-bengioy-ad/data/curated/list_datasets_cc.sh"
+
+weights
+-------
+
+``weights`` contains curated models weights to the benefit of the Mila community.
+To request the addition of a dataset or a preprocessed dataset you think could
```
btravouillon, line 140: Should we use `dataset` or `weights`?
```suggestion
To request the addition of a weight or a preprocessed weight you think could
```

Seems weird to me, but if we keep `dataset` this may lead to some confusion.

docs/Information_storage.rst
```diff
@@ -131,3 +132,19 @@ following command:
...
+benefit the research of others, you can fill `this form
+<https://forms.gle/HLeBkJBozjC3jG2D9>`_.
+
+Weights in ``datasets/restricted`` are restricted and require an explicit
```
btravouillon, line 144: ?
```suggestion
Weights in ``weights/restricted`` are restricted and require an explicit
```

btravouillon (changes requested):

btravouillon (approved):
Looks good to me. 👍

obilaniu (approved):

PR #200, opened 2023-07-05
Restructure the minimal examples section [DOC-144]

lebrice:
Les autres PR d'exemples minimalistes devraient être rebase sur ces changements (ou bien merge ces commits). Cette PR-là change un peu la structure de la section d'exemples minimalistes:
- le nom des fichiers passe de _index.rst --> index.rst
- Les exemples ne se trouvent plus tous dans la même page
- Améliore l'auto-génération des README.rst pour voir les exemples sur GitHub:
    - Ajoute un header dans le fichier README.rst auto-généré pour dire que le contenu ne devrait pas être modifié, et indiquer la source
    - Force chaque exemple a contenir un lien vers le code source sur GitHub.
    - Fix l'indentation à trois espaces


docs/examples/distributed/multi_node/index.rst
```diff
@@ -1,12 +1,12 @@
-003 - Multi-Node (DDP) Job
-=====================================
+Multi-Node (DDP) Job
+********************
```
satyaog, line 2:
```suggestion
====================
```

docs/examples/distributed/single_gpu/index.rst
```diff
@@ -1,26 +1,28 @@
-001 - Single GPU Job
-====================
+.. _single_gpu_job:
+
+Single GPU Job
+**************
```
satyaog, line 4:
```suggestion
==============
```

docs/examples/good_practices/index.rst
```diff
@@ -1,5 +1,13 @@
-**************
-Good Practices
-**************
+*******************
+Good practices
+*******************
```
satyaog, line 3:
```suggestion
==============
Good Practices
==============
```

btw, I think it would be ok to have only checkpointing alone so when we click on it in the left pane we land on the example directly

docs/examples/good_practices/index.rst
```diff
@@ -1,5 +1,13 @@
...
 
-.. include:: /examples/good_practices/checkpointing/_index.rst
+Work-in-progress: This section will contain minimal examples illustrating various good practices
+that should be observed when using the Mila cluster.
```
satyaog, line 6: Work-in-progress: This section will contain minimal examples illustrating
various good practices that should be observed when using the Mila cluster.

docs/examples/frameworks/index.rst
```diff
@@ -2,5 +2,14 @@
 Software Frameworks
 *******************
 
+(Work-in-progress): This section shows how to setup a development environment for various
+frameworks.
```
satyaog, line 6:
```suggestion
(Work-in-progress): This section shows how to setup a development environment
for various frameworks.
```

docs/examples/frameworks/pytorch_setup/README.rst
```diff
@@ -1,77 +1,88 @@
...
-   #SBATCH --cpus-per-task=1
-   #SBATCH --mem=16G
-   #SBATCH --time=00:15:00
-   #SBATCH --partition=unkillable
+    #!/bin/bash
+    #SBATCH --gres=gpu:1
+    #SBATCH --cpus-per-task=1
+    #SBATCH --mem=16G
+    #SBATCH --time=00:15:00
+    #SBATCH --partition=unkillable
```
satyaog, line 34: Are we changing to 4 spaces? We should probably have single huge commit changing all 3 spaces to 4 spaces. Else, we should stick to 3 spaces

docs/examples/distributed/index.rst
```diff
@@ -2,6 +2,11 @@
...
-.. include:: /examples/distributed/002_multi_gpu/_index.rst
-.. include:: /examples/distributed/003_multi_node/_index.rst
+.. toctree::
+    :maxdepth: 1
+    :glob:
+    :numbered:
+
+    single_gpu/index
+    multi_gpu/index
+    multi_node/index
```
satyaog, line 12:
```suggestion
    */index
```

docs/examples/frameworks/pytorch_setup/index.rst
```diff
@@ -5,22 +5,29 @@ PyTorch Setup
...
+repository.
+<https://github.com/mila-iqia/mila-docs/tree/master/docs/examples/frameworks/pytorch_setup>`_
+
+
 * :ref:`Quick Start`
 * :ref:`Running your code`
 * :ref:`Conda`
 
 
+
```
satyaog, line 19:
```suggestion
```

satyaog (changes requested):
If you don't agree with the suggestions I'll change my review to approved ;)

(you can batch all the changes you would like in a single commit btw)

docs/Minimal_examples.rst
```diff
@@ -1,8 +1,23 @@
-.. ****************
-.. Minimal Examples
-.. ****************
+****************
+Minimal Examples
+****************
 
+This section contains some minimal examples of how to run jobs on the Mila cluster.
```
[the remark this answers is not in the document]
lebrice: I prefer "examples of" than "examples about"

docs/Minimal_examples.rst
```diff
@@ -1,8 +1,23 @@
...
-.. include:: examples/distributed/index.rst
-.. include:: examples/good_practices/index.rst
+Each example is self-contained and can be run as-is directly on the cluster without error.
+Examples has the following structured:
+
+* ``job.sh``: SLURM ``sbatch`` script. Can be launched with ``sbatch job.sh``.
+* ``main.py``: Example python script.
+
+Some examples (for example :doc:`this one <examples/distributed/multi_gpu/_index>`) are displayed in terms of the difference
+between some "base" example and the "new" example.
```
lebrice, line 14:
```suggestion
Some examples are displayed as a difference with respect to a "base" example. For instance, the :doc:`multi-gpu example <examples/distributed/multi_gpu/index>` is shown as a difference with respect to the :doc:`single-gpu example <examples/distributed/single_gpu/index>`.
```

docs/Minimal_examples.rst
```diff
@@ -1,8 +1,23 @@
...
+
+
+
+.. toctree::
+    :maxdepth: 1
+
+    examples/frameworks/index
+    examples/distributed/index
+    examples/good_practices/index
```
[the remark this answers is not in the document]
lebrice: Interesting idea, I'll try it out.

docs/examples/distributed/index.rst
lebrice: The ordering of the examples is important here (they are numbered), so I can't use the glob this way.

docs/examples/frameworks/pytorch_setup/README.rst
lebrice: 4 spaces. And I'd be in favour of merging the PR as-is, if you don't mind.

docs/examples/good_practices/index.rst
lebrice: We're going to add more examples in the "good practices" section soon. It's fine IMO if the section only has one page for now.

docs/examples/good_practices/index.rst
lebrice: ?

satyaog: oups, missing some ticks

docs/examples/distributed/index.rst
satyaog: ah yes you're right

docs/examples/frameworks/pytorch_setup/README.rst
satyaog: Sounds like a one man decision ;). I would argue that style changes should be done project wide but if we're eager to merge this one I suppose we can take a shortcut here. I would still prefer for this kind of style modification, which is very subjective, to be done in a separate PR in the future. (also, not talking about python code indentation, for which 4 is the standard, only for rst, for which 3 spaces is used pretty much everywhere)

satyaog: Added an issue to resolve this question properly : https://github.com/mila-iqia/mila-docs/issues/208

lebrice: The fixed requirement here is that all code blocks in the examples (.diffs or not) need to be directly pasteable into an editor and land correctly indented (4 spaces).
I'm fine with whatever solution that addresses this requirement.

docs/examples/preprocess.py
```diff
@@ -31,36 +82,103 @@ def preprocess():
...
+                line = line.strip(" ")
+                if line.startswith(":language:"):
+                    lang = line[len(":language:") :].strip(" ")
+                elif line.startswith(".. literalinclude:: ") or not line:
+                    break
+            del content_lines[i : i + 1 + j]
+
+            insert = [f".. code:: {lang}", ""] + [
+                f"    {_l}".rstrip(" ") for _l in (docs_root / path).read_text().split("\n")
+            ]
```
satyaog, line 162:
```suggestion
            insert = [f".. code:: {lang}", ""] + [
                f"   {_l}".rstrip(" ") for _l in (docs_root / path).read_text().split("\n")
            ]
```

lebrice: Fixed in 918355e155cebebff110a16fa9df69666f6ef246

satyaog (approved):

docs/Minimal_examples.rst
```diff
@@ -1,8 +1,22 @@
...
+Minimal Examples
+****************
 
+This section contains some minimal examples of how to run jobs on the Mila cluster.
 
-.. include:: examples/frameworks/index.rst
-.. include:: examples/distributed/index.rst
-.. include:: examples/good_practices/index.rst
+Each example is self-contained and can be run as-is directly on the cluster without error.
+Examples has the following structured:
```
lebrice, line 8:
```suggestion
Each example has the following structure:
```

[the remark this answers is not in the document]
lebrice: Thanks @soline-b , fixed in b65f07910eed917294fdcbad249f8ee728044d70

docs/Minimal_examples.rst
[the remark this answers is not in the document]
lebrice: I'm unable to make this work locally. Perhaps I'm doing something wrong. Did you mean to remove the toctree here instead have something like this in the `index.rst`?

```
.. toctree::
   :caption: Minimal Examples
   :maxdepth: 2

   Minimal_examples
   examples/frameworks/index
   examples/distributed/index
   examples/good_practices/index
```


If so, this doesn't seem to work correctly, there is still only "Minimal Examples" under the "MINIMAL EXAMPLES" section in the left navbar.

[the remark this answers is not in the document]
lebrice: Oh also: I think that using `.. include` would make them all end up in the same page, right? I ideally wouldn't want that to happen here.

docs/examples/good_practices/index.rst
lebrice: Fixed in 65d6157897f45be881f5ce33dc1e99eb5bf817fc

docs/examples/good_practices/index.rst
lebrice: Fixed in 65d6157897f45be881f5ce33dc1e99eb5bf817fc

docs/examples/preprocess.py
```diff
@@ -31,36 +82,107 @@ def preprocess():
...
+                    f"literalinclude: {path}"
+                )
+            lang = ""
+            j = 0
+            for j, line in enumerate(content_lines[i + 1 :]):
+                line = line.strip(" ")
+                if line.startswith(":language:"):
+                    lang = line[len(":language:") :].strip(" ")
+                elif line.startswith(".. literalinclude:: ") or not line:
+                    break
```
breuleux, line 157: This seems a bit brittle as a way to figure out where the block ends, compared to simply checking if the (non-stripped) line starts with a non-space character.

docs/examples/preprocess.py
```diff
@@ -31,36 +82,107 @@ def preprocess():
...
+            del content_lines[i : i + 1 + j]
```
breuleux, line 158: Another strategy to remove the literalinclude would be to change `.. literalinclude: ...` to `..`, which would be interpreted as a comment block. If you then put the insert on top, you don't actually need to know where the block ends.

docs/examples/generate_diffs.sh
```diff
@@ -24,19 +24,19 @@ generate_diff() {
...
-generate_diff distributed/001_single_gpu/job.sh good_practices/checkpointing/job.sh
-generate_diff distributed/001_single_gpu/main.py good_practices/checkpointing/main.py
+generate_diff distributed/single_gpu/job.sh good_practices/checkpointing/job.sh
+generate_diff distributed/single_gpu/main.py good_practices/checkpointing/main.py
 
 # single_gpu -> hpo_with_orion
-generate_diff distributed/001_single_gpu/job.sh good_practices/hpo_with_orion/job.sh
-generate_diff distributed/001_single_gpu/main.py good_practices/hpo_with_orion/main.py
+generate_diff distributed/single_gpu/job.sh good_practices/hpo_with_orion/job.sh
+generate_diff distributed/single_gpu/main.py good_practices/hpo_with_orion/main.py
```
breuleux, line 40: Could we loop over `**/job.sh` and `**/main.py` instead, so we don't have to remember to update this file whenever we add a new example?

docs/examples/preprocess.py
```diff
@@ -31,36 +82,107 @@ def preprocess():
...
+
+            insert = [f".. code:: {lang}", ""] + [
+                # NOTE: The code block in the .rst files is indented with 3 spaces, but the
+                # rendered code blocks in the README.rst on GitHub and in the online docs have the
+                # correct indented for a Python script (4 spaces).
+                f"   {_l}".rstrip(" ")
+                for _l in (docs_root / path).read_text().split("\n")
+            ]
+            content_lines = content_lines[:i] + insert + content_lines[i + 1 :]
```
breuleux, line 167: This drops `content_lines[i]`, is this what we want? We already removed the literalinclude lines.

docs/examples/generate_diffs.sh
lebrice: Perhaps I don't understand, but here we need to define the from-to relationship between examples. How do you suggest we do that automatically?

docs/examples/preprocess.py
lebrice: Fixed in 34cc66ec8cb48041ee495816412a124f5ea17a8f

docs/examples/generate_diffs.sh
breuleux: Ah yeah I had a brain fart here, sorry.

Although, one way to do this automatically would be to represent the examples as a Git arborescence: you write the code for single_gpu, then you create a multi_gpu branch that has one commit that is the difference, and so on, each example being its own branch. Then you can run a series of checkouts and cp to populate the examples directory in the actual docs repo. With this method you can also update the whole set of examples more easily by changing the base and looping on git rebase. But in any case this is out of scope for this PR.

docs/examples/preprocess.py
lebrice: Interesting idea. I didn't go for that approach, but I did simplify the code quite a bit in 
34cc66ec8cb48041ee495816412a124f5ea17a8f :

Now it works essentially like this:

```python
old_content_lines = old_content.splitlines() 
new_content: list[str] = [".. This file was generated blablabla", ""]

current_line_index = 0
while line_index < len(old_content_lines)
     line = old_content_lines[line_index]
     if not is_literalinclude_block(line):
         new_content.append(line)
         current_line_index += 1
         continue
     language_line = old_content_lines[line_index+1]
     path: Path = _get_path(line)
     language: str = _get_language(language_line)
     new_content += path.read_text().splitlines()
    current_line_index += 2
```

breuleux: Yeah, that works, although we need to make sure the literalinclude statements are exactly two lines long.

docs/examples/preprocess.py
lebrice: Fixed in 4bc0bae2107cbd72e270157da463d4d44458b2eb

docs/examples/generate_diffs.sh
lebrice: Yeah that's interesting!

I also played around with the idea of us editing the .diff files directly instead of the example files, and re-generating the derived examples if something changes in the base.

For now, with this limited number example dependencies, it's pretty manageable to make a change in the base example and then copy-paste it to the 5-6 derived examples. However there is potentially a better solution in the long run for sure.
I agree with you, I don't think it's worth it to spend too much time working on that right now.

docs/examples/preprocess.py
lebrice: Fixed in 83fcb30b5673956fc791f5af08ea2d6d35252943

docs/Minimal_examples.rst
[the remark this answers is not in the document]
lebrice: Fixed in 0f5ddd05b40561873df6f3bc566bffce36dedced

docs/examples/preprocess.py
```diff
@@ -31,36 +86,170 @@ def preprocess():
...
+
+        def _get_block_length() -> int:
+            length = 1
+            while (
+                line_index + length < len(content_lines)
+                and content_lines[line_index + length]
+                and not content_lines[line_index + length].isspace()
+            ):
+                length += 1
+            return length
```
breuleux, line 198: I haven't tested so maybe I'm following the code wrong, but you only stop on empty/whitespace lines, so it seems to me that the function will return 4 on the following block:

```rst
.. literalinclude:: x.py
  :language: python
.. important::
  Don't run vscode on the login nodes you guys.
```

And 1 on this one, which I'm not certain is legal or not:

```rst
.. literalinclude:: x.py

  :language: python
```

lebrice: Fixed here: 806c59da5ecc90e3393aead927628820d90ea94d

breuleux (approved):

PR #210, opened 2023-08-11
Add info that ARCHIVE is mounted on CPU-only nodes

btravouillon:


lebrice (approved):
A quick rant:
I would love for these `tar` commands (and other commands in the docs) to use the `--long-form` for the arguments, instead of one-letter command arguments that just feel like incomprehensible bash magic.

satyaog (approved):

PR #214, opened 2023-09-11
load anaconda

Gregpy:
Loading anaconda added to environment creation instructions. I also would think it may be preferable to name the environment something besides 'pytorch'. Something more unique like 'pytorch_ex'. 
I didn't  change that though. Also, maybe the python version could be updated somewhat (though this could lead to changes in other parts of examples).


lebrice (approved):

satyaog (approved):

satyaog:
Thanks a lot @Gregpy !

PR #227, opened 2023-10-16
errata and email

gyom:
I'd added the email for the mailing group `idt.cheatsheet@mila.quebec`, which was tested and works as expected.

I've added some information in the errata section because, after talking with Bruno and Olexa, I realized that I wasn't describing preemption accurately. In fact, I was make wrong statements. I don't need to explain everything on the cheat sheet, but I least I don't want to suggest incorrect interpretations.

I've added some clarification about who has access to the "default" accounts for professors. Again, the cheat sheet was oversimplifying things in a way that could lead to incorrect interpretations.

Adding a reference to a typo also.

### DIFF
--- a/docs/Cheatsheet.rst
+++ b/docs/Cheatsheet.rst
@@ -36,11 +36,10 @@ The jobs from the `long` partition can be preempted to allow jobs in the `main` 
 but jobs in `main` are never going to be preempted to allow for other jobs in `main`, no matter how much
 of the "fair use" a user has already consumed.
 
-This is why it can be considered rude (or just bad practice) to queue a large number of jobs
-on `main` at midnight on the Mila cluster with a lot of free capacity.
-On the next day, those jobs might still be running and much of the compute capacity
-will be shared very unequally between the users, even though it seemed like a good strategy
-to schedule jobs when the cluster was idle.
+Jobs for the `unkillable` partition can preempt everything except other jobs on the `unkillable` partition.
+
+A good explanation of priorities and preemption will be added to the documentation,
+but the cheat sheet (or the cheat sheet errata) is not the place for this.
 
 Default accounts on DRAC oversimplified
 ---------------------------------------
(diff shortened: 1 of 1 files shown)


docs/Cheatsheet.rst
```diff
@@ -19,8 +19,44 @@ which is the original source of information.
...
+Partition preemption is not explained accurately on page 2
+----------------------------------------------------------
+
+The preemption on the Mila cluster is a bit more complicated than what is described in the cheat sheet.
+The jobs from the `long` partition can be preempted to allow jobs in the `main` partition to run,
+but jobs in `main` are never going to be preempted to allow for other jobs in `main`, no matter how much
+of the "fair use" a user has already consumed.
+
+This is why it can be considered rude (or just bad practice) to queue a large number of jobs
+on `main` at midnight on the Mila cluster with a lot of free capacity.
```
obilaniu, line 40: It is not possible to run more than two `main` jobs in parallel per user, because of the requirement of 1+ GPU per GPU job (obvious) and maximum 2 GPUs used in total by jobs inside the `main` partition.

I'd stick to `long` for the explanation and say that `unkillable` jobs can preempt `main`, `main` can preempt `long`, but like never preempts like.

gyom: Oh, wow, so that "max resource usage" is a total among all the jobs in that partition, and not just a per-job maximum ...

Then we're kind of back to square one in that people are going to need to submit the bulk of their stuff to `long` and hope that it runs there, and those resources are going to get hogged by someone who submitted 100 jobs yesterday when the cluster was idle.

It's just that everyone gets a 2 GPU from `main` in order to declog the cluster slowly, and those will preempt the jobs from whoever submitted the 100 jobs to `long`. Is that it?

gyom: I've modified my description to be less thorough (and thus less wrong). I think it's more reasonable this way, and I will also open another ticket here to represent the fact that we should document this partition thing better on docs.mila.quebec.

obilaniu (approved):

btravouillon (approved):
Looks good to me. 👍

ahmam (approved):

PR #232, opened 2023-10-30
Make @lebrice, @satyaog, @obilaniu code owners of docs/examples

lebrice:


btravouillon (approved):
Looks good to me. 👍

CODEOWNERS
```diff
@@ -1,2 +1,2 @@
 *        @btravouillon @satyaog
-
+docs/examples/        @lebrice @satyaog @obilaniu
```
satyaog, line 2: Do I need to be there if I'm already in the top line?

lebrice: Yeah, according to https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file :
```
# These owners will be the default owners for everything in
# the repo. Unless a later match takes precedence,
# @global-owner1 and @global-owner2 will be requested for
# review when someone opens a pull request.
*       @global-owner1 @global-owner2

# Order is important; the last matching pattern takes the most
# precedence. When someone opens a pull request that only
# modifies JS files, only @js-owner and not the global
# owner(s) will be requested for a review.
*.js    @js-owner #This is an inline comment.
```

satyaog (approved):

PR #236, opened 2023-11-07
Exclude MIG in multi-gpu/node examples (DOC-174)

lebrice:


lebrice:
Unnecessary, since the GPU type is already set in the --gpus flag

PR #244, opened 2024-05-16
updating the cheatsheet section with the new pdf

gyom:
New pdf, plus some description in the appropriate page.


gyom:
I think I'd like to keep the previous versions in the pdf for the cheatsheet in Github repo. We don't need to link to them, but it would still be nice to keep them there in case we want to add a collection of links to all the previous versions at some point.

That's just my explanation as to why I didn't delete the previous pdf in my PR despite not including any link to it. It's just an opinion, but I'm not against removing the old pdf from the main brain if someone wants to make an argument for that.

btravouillon (approved):
Looks good to me. 👍

PR #245, opened 2024-06-20
Bugfix GPU ordinal computation

obilaniu:
With `--gpus-per-task=rtx8000:1`, all tasks only see `1` GPU. Therefore, the only valid GPU ordinal is `0`, even if each task sees a different single GPU.

For extra robustness, use the common trick of calculating the ordinal as `ordinal = rank % device_count()`, which works both in and outside of SLURM.

### DIFF
--- a/docs/examples/distributed/multi_gpu/README.rst
+++ b/docs/examples/distributed/multi_gpu/README.rst
@@ -29,6 +29,7 @@ Click here to see `the code for this example
     #SBATCH --cpus-per-task=4
    -#SBATCH --ntasks-per-node=1
    +#SBATCH --ntasks-per-node=4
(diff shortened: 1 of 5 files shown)


lebrice (approved):
Thanks!

By the way, the reason the "generate files" step in the CI is failing is because the examples are build from the source code (more precisely, the diff between the base example and the current example),  so we use pre-commit to make sure that the .rst files are regenerated properly from the source code.

You'd normally setup pre-commit via `pre-commit install`, and then it would run the generate_files script after each commit. To get the same effect now, you'd run `pre-commit run --all-files` and commit the changes.

lebrice (approved):

btravouillon (approved):
Looks good to me. 👍

PR #246, opened 2024-06-21
Defensive improvements to ACL instructions for inattentive users.

obilaniu:
Prevent copy-pasting with undefined `${USER2}` variable by using POSIX shell parameter expansion feature "`${parameter:?word}`", which prints an error "word" if `$parameter` is empty or undefined.

In the context of `setfacl [-args] user:${USER2}:x`, an empty string is legal and indicates "the owner, whoever that is". More than one user has removed their own permissions by `setfacl`'ing themselves out of them accidentally. It's time to put a safety on this foot-gun.

Also move a note from Step 3 to below Step 4 in order to make Step 4 more prominent and less easily-skipped by those in a hurry.

### DIFF
--- a/docs/Userguide_sharing_data.rst
+++ b/docs/Userguide_sharing_data.rst
@@ -51,7 +51,7 @@ search and delete each others' files:
 
 .. code-block:: bash
 
-   setfacl -Rdm user:${USER2}:rwx $SCRATCH/X/Y/Z/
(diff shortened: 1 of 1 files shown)


btravouillon (approved):
Looks good to me. 👍

lebrice (approved):

PR #251, opened 2024-08-07
Change `conda clean --all` to `-it`

obilaniu:
Due to legacy circumstances, `conda` was falling back to breakable symlinks rather than using safer hardlinks when "copying" code from its package cache. When `conda` selects symlinks, a `clean --all` can damage irretrievably a conda env. Avoid this by only cleaning the index and tarballs, but not the unpacked files that are potentially the targets of said symlinks.


satyaog (approved):

btravouillon (approved):
Looks good to me. 👍

PR #255, opened 2024-09-12
Zsh

satyaog:


docs/Userguide_login.rst
```diff
@@ -189,3 +189,11 @@ require a lot of resources: editing via VSCode, building conda environments, tes
...
 While it is possible to tweak it to allocate a GPU, doing so will prevent simultaneous
 connections to it (until Slurm is upgraded to version 22.05 or later).
+
+Using a non-Bash Unix shell
+---------------------------
+
+While IT and Infra will not provide personalized support for non-Bash Unix
+shell on the cluster, the cluster is designed to support both Bash and Zsh
+shells. If you think things should work with Zsh and they don't, please contact
+`Mila's IT support <https://it-support.mila.quebec>`_.
```
satyaog, line 199: Not sure if I should include details on what is implied by "the cluster is designed to support both Bash and Zsh". I'm thinking of how the modules are loaded but I'm not sure if this should be mentioned and if there is more.

satyaog, line 199: Not sure if this should be rephrased, I would like it and infra to give me some feedback as they would be directly involved in answering user

docs/Userguide_login.rst
```diff
@@ -189,3 +189,11 @@ require a lot of resources: editing via VSCode, building conda environments, tes
 The ``mila-cpu`` entry should be in your ``.ssh/config``. Changes are at your own risk.
 While it is possible to tweak it to allocate a GPU, doing so will prevent simultaneous
 connections to it (until Slurm is upgraded to version 22.05 or later).
+
+Using a non-Bash Unix shell
+---------------------------
+
+While IT and Infra will not provide personalized support for non-Bash Unix
```
satyaog, line 196: Would `1-on-1 support` be better? The idea is that users would not request help to setup their stuff with Zsh, but still know that it should work and for the team to know if things break

docs/Userguide_login.rst
btravouillon: I would got with Mila.
```suggestion
While Mila will not provide personalized support for non-Bash Unix
shell on the cluster, the cluster is designed to support both Bash and Zsh
shells. If you think things should work with Zsh and they don't, please contact
`Mila's IT support <https://it-support.mila.quebec>`_.
```

docs/Userguide_login.rst
satyaog: Message updated

docs/Userguide_login.rst
```diff
@@ -189,3 +189,11 @@ require a lot of resources: editing via VSCode, building conda environments, tes
...
 While it is possible to tweak it to allocate a GPU, doing so will prevent simultaneous
 connections to it (until Slurm is upgraded to version 22.05 or later).
+
+Using a non-Bash Unix shell
+---------------------------
+
+While Mila does not provide support in debugging your shell setup, Bash is the
+standard shell to be used on the cluster and the cluster is designed to support
+both Bash and Zsh shells. If you think things should work with Zsh and they
+don't, please contact `Mila's IT support <https://it-support.mila.quebec>`_.
```
satyaog, line 199: Not sure if I should include details on what is implied by "the cluster is designed to support both Bash and Zsh". I'm thinking of how the modules are loaded but I'm not sure if this should be mentioned and if there is more.

btravouillon (approved):
Looks good to me. 👍

priscillaben (approved):

PR #267, opened 2025-03-04
Update information on filesystems quotas

obilaniu:
The limit on `$SCRATCH` was lowered to 10TiB.

### DIFF
--- a/docs/Information_storage.rst
+++ b/docs/Information_storage.rst
@@ -12,10 +12,10 @@ Path                                                   Performance Usage        
 ``/network/weights/``                                  High        * Curated models weights (read only)
 ``$HOME`` or ``/home/mila/<u>/<username>/``            Low         * Personal user space                  100GB/1000K         Daily  no
                                                                    * Specific libraries, code, binaries
-``$SCRATCH`` or ``/network/scratch/<u>/<username>/``   High        * Temporary job results                20TB/no             no     90 days
+``$SCRATCH`` or ``/network/scratch/<u>/<username>/``   High        * Temporary job results                10TB/no             no     90 days
                                                                    * Processed datasets
                                                                    * Optimized for small Files
-``$SLURM_TMPDIR``                                      Highest     * High speed disk for temporary job    4TB/-               no     at job end
+``$SLURM_TMPDIR``                                      Highest     * High speed disk for temporary job    no/no               no     at job end
                                                                      results
 ``/network/projects/<groupname>/``                     Fair        * Shared space to facilitate           1TB/1000K           Daily  no
                                                                      collaboration between researchers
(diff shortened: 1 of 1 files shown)


docs/Information_storage.rst
```diff
@@ -12,10 +12,10 @@ Path                                                   Performance Usage
 ``/network/weights/``                                  High        * Curated models weights (read only)
 ``$HOME`` or ``/home/mila/<u>/<username>/``            Low         * Personal user space                  100GB/1000K         Daily  no
                                                                    * Specific libraries, code, binaries
-``$SCRATCH`` or ``/network/scratch/<u>/<username>/``   High        * Temporary job results                20TB/no             no     90 days
+``$SCRATCH`` or ``/network/scratch/<u>/<username>/``   High        * Temporary job results                10TB/no             no     90 days
                                                                    * Processed datasets
                                                                    * Optimized for small Files
-``$SLURM_TMPDIR``                                      Highest     * High speed disk for temporary job    4TB/-               no     at job end
+``$SLURM_TMPDIR``                                      Highest     * High speed disk for temporary job    4TB/234.4M          no     at job end
```
btravouillon, line 18: There is no quota on the tmp dir. I can't say how we defined 4TB in the documentation. I guess you should remove both values.

btravouillon (changes requested):

docs/Information_storage.rst
obilaniu: I can assure you that there is an ultimate limit of 234.4M inodes on `$SLURM_TMPDIR` because it is (still) formatted `ext4`, and `ext4` filesystems are created with a large but fixed-size inode table. `stat -f /var/opt/slurm/localstorage` verifies it is 234.4M on `cn-c`, `cn-g` and `cn-l` nodes, comprising the bulk of the cluster, and I suspect the same applies to all nodes except perhaps the DGXes.

And as unbelievable as it sounds, there has been a case only three months ago (Dec. 18, 2024) where somebody was unpacking a tarball with 72.5M files on our cluster and if four jobs were allocated on the same node, the fourth one would fill the inode table (4x72.5M = 290M > 234.4M).

The 4TB is arguably incorrect, it should be 3.5 TiB. But it comes from the smallest nodes' `TmpDisk` as declared in `slurm.conf`. The `cn-c` are declared to have 3,000,000M (MiB), and all other nodes have 3,600,000 or more. But `lsblk` says that the `cn-c` have a disk 3840755982336 bytes in size, and that is approximately 3,662,830 MiB or very nearly 3.5 TiB.

btravouillon: It's not a quota, it's a disk limit. These limits can change depending on the nodes. For example, DGX nodes have 30TB of tmp disk. No 4TB quota, a single user can use 10TB on the node.

btravouillon (approved):
Looks good to me. 👍

ahmam (approved):

PR #271, opened 2025-04-16
ci: Use ubuntu-latest runner

btravouillon:
Ubuntu 20.04 is deprecated. Using the latest ubuntu should be fine for this project, no need to pin a specific release.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   lint_docs:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


obilaniu (commented):
Maybe 24.04? It might a bit "dangerous" to let `-latest` suddenly break things from under us suddenly.

btravouillon:
> Maybe 24.04? It might a bit "dangerous" to let `-latest` suddenly break things from under us.

That was my first edit, then I noticed 2 others CI files already use `ubuntu-latest`. It would indeed be safer to pin the release. That being said I wonder if we won't hit a similar issue than today: we pinned 20.04 a long time ago and discover now that it is deprecated. Having to upgrade now could break things.

One the other hand, `ubuntu-latest` should change every two years for each new LTS release. This would mean if something should break in `ubuntu-26.04`, we could still pin `ubuntu-24.04` and give us some time to analyze the root cause of the failures.

obilaniu (approved):

ahmam (approved):

PR #273, opened 2025-05-13
Removed mention of chatbot

gyom:
It was about time that we removed this section from the documentation.


btravouillon (approved):
Looks good to me. 👍

obilaniu (approved):

PR #291, opened 2025-10-31
Update DRAC allocation information

obilaniu:
Also drop mention of old clusters.


bouthilx:
The focus will still be on Narval with mentions of new A100: https://docs.mila.quebec/Extra_compute.html#narval.

bouthilx:
Swapping Narval with Rorqual should probably require only minor changes.

satyaog (approved):

PR #310, opened 2026-02-06
Fix Minimal Examples/Software Frameworks mkdocs pages

satyaog:
* Fix the markdown documentation files of the Minimal Examples/Software Frameworks section (https://mila-mkdocs--310.org.readthedocs.build/en/310/examples/frameworks/)
* Fix the section index for Minimal Examples/Software Frameworks
  * Clicking on Minimal Examples/Software Frameworks in the navigation bar will open the index
  * The sub-navigation menues of Minimal Examples/Software Frameworks will not duplicate Software Frameworks as the first entry and the navigation section


.cursor/skills/rst-to-md-mkdocs-one-page/SKILL.md
lebrice: Same as other PRs. This doesn't belong here IMO.

docs/examples/advanced/imagenet/README.md
```diff
@@ -1,21 +1,25 @@
+<!-- NOTE: This file is auto-generated from examples/advanced/imagenet/index.md
+     This is done so this file can be easily viewed from the GitHub UI.
+     DO NOT EDIT -->
+
 ### Multi-Node / Multi-GPU ImageNet Training
```
lebrice: Headings should be generated with heading level 1.

docs/examples/preprocess.py
```diff
@@ -236,6 +250,100 @@ def _get_path_to_file(filename: str) -> Path:
...
 
 
+def _get_path_to_file_md(filename: str, example_readme_template_path: Path) -> Path:
+    """Resolve snippet path for Markdown: from project root (docs/...) or relative."""
+    if filename.startswith("docs/"):
+        return DOCS_ROOT.parent / filename
+    if "/" not in filename:
+        return example_readme_template_path.parent / filename
+    raise NotImplementedError(
+        f"Example {example_readme_template_path} uses a path in --8<-- that is neither "
```
lebrice: Idk if it's useful to know, but just FYI the include-markdown plugin raises an error when a file is missing. Does the Snippets plugin do the same? Maybe we wouldn't need to check here then.

mkdocs.yml
```diff
@@ -13,7 +13,7 @@ theme:
     - navigation.sections
     - navigation.expand
     - navigation.path
-    # - navigation.indexes
+    - navigation.indexes
```
lebrice: Did it work in your case for the examples? I had a lot of trouble with this before, in particular because of the interaction with literate-nav. 

I guess I'll just look at the preview and see for myself!

scripts/rst_to_mkdocs_md.py
```diff
@@ -66,7 +66,7 @@ def repl_raw(m: re.Match) -> str:
     def repl_include(m: re.Match) -> str:
         inc = m.group(1).strip()
         inc_md = re.sub(r"\.rst$", ".md", inc)
-        return f"<!-- include: {inc_md} -->\n\n"
+        return f'--8<-- "{inc_md}"\n\n'
```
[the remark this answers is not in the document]
lebrice: Actually no, it's really not that necessary. Whatever you prefer is fine by me. We can change things later. Sorry.

mkdocs.yml
satyaog: Ah but didn't you specifically pointed towards the using of `navigation.indexes` in your comment here?

https://github.com/mila-iqia/mila-docs/pull/313#issuecomment-3872777009

Let me know if I understood incorrectly. But yes it does fix the index issue there was in the navigation bar of the section

satyaog:
Please notice the particularity here which the work on the preprocessing which touches all the minimal examples files, corrected or not.

So the choice would be to have a big PR with all the minimal examples files being fixed (so merging together the 4 open PRs), or keep smaller as independent as possible PRs. I would prefer less files to review personally.

Please also note that the base branch is itself a work in progress branch with broken sections all over the place, not the master branch, so it should be reviewed in its entirety before merging regardless of the approved PRs we're doing now. We should probably review them in a more relaxed way until we have something stable.

PR #317, opened 2026-02-09
Don't auto-expand the docs sections

lebrice:
Don't auto- expand until maybe after we reorganize the content a bit.
Atm it's too many pages all at once, and doesn't match the sphinx structure.

Signed-off-by: Fabrice Normandin <normandf@mila.quebec>


lebrice:
@satyaog  it doesn't fix the rendering of a broken page, it just makes the doc not fully-expanded by default (to match the current sphinx doc)

One example is the "Computational infrastructure and policies" page.

PR #319, opened 2026-02-10
Fix header levels in pages of "Extras" section

lebrice:
Signed-off-by: Fabrice Normandin <normandf@mila.quebec>

### DIFF
--- a/docs/VSCode.md
+++ b/docs/VSCode.md
@@ -1,5 +1,4 @@
-
-### Visual Studio Code
+# Visual Studio Code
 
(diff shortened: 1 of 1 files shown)


lebrice:
> I think that the file VSCode.md have to be its headers fixed too. (I don't want to interfere in your work but I can do it if you want)
> 
> Approved for the other files!

Good catch @soline-b ! Fixed in 5b8c409e5ce3e45396f45a17a3b0ad2092a0e80a

PR #353, opened 2026-03-20
docs(examples): add GPU model alternatives and troubleshooting tip

wietzesuijker:
Add commented alternatives (`rtx8000`, any GPU) to example job scripts so users learn they can specify GPU models. Add a troubleshooting tip to the quickstart page for the `ReqNodeNotAvail` error.

Based on review feedback, keeps `l40s:1` as the default (appropriate size for example tasks). Multi-node and job array examples omit the "any GPU" option to avoid mixed-model issues.

### DIFF
--- a/docs/examples/good_practices/slurm_job_arrays/job.sh
+++ b/docs/examples/good_practices/slurm_job_arrays/job.sh
@@ -3,6 +3,7 @@
 #SBATCH --ntasks-per-node=1
 #SBATCH --cpus-per-task=4
 #SBATCH --gpus-per-task=l40s:1
+## Or --gpus-per-task=rtx8000:1    to request a different GPU model
 #SBATCH --mem-per-gpu=16G
 #SBATCH --time=00:15:00
 
--- a/docs/Userguide_quick_start_train_first_model.md
+++ b/docs/Userguide_quick_start_train_first_model.md
@@ -117,6 +117,12 @@ Submitted batch job 8888888
 
 You will see something like `Submitted batch job 8888888`. Note the job ID.
 
+!!! tip "Job stuck with `ReqNodeNotAvail`?"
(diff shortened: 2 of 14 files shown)


docs/examples/good_practices/slurm_job_arrays/job.sh
```diff
@@ -2,7 +2,8 @@
 #SBATCH --ntasks=1
 #SBATCH --ntasks-per-node=1
 #SBATCH --cpus-per-task=4
-#SBATCH --gpus-per-task=l40s:1
+#SBATCH --gpus-per-task=1
+# Or use --gpus-per-task=rtx8000:1 to request a specific GPU model.
```
satyaog:
```suggestion
#SBATCH --gpus-per-task=l40s:1
## Or --gpus-per-task=rtx8000:1   to request a different GPU model
## Or --gpus-per-task=1           for any GPU model
```

satyaog (changes requested):
Thanks for your feedback and contribution to the docs!

We do like the `l40s` as its a rather small GPU which fits well the task of testing an example (not doing meaningful work). Having users land on big GPUs for these minimal examples would be a waste of resources we are unsure we want right now. As such, we would suggest the following changes, except for the multi node example as Slurm could give the user different GPU models for the same task which could lead to unexpected behaviours. The same could apply to the Slurm job array example, different GPU models for the same job could lead to unexpected behaviours, particularly when reproducibility is important.

Let us know what you think. I believe the ideal solution would be to avoid mentioning `## Or --gpus-per-task=1            for any GPU model` in the minimal examples (it could be mentioned outside of the minimal examples) but if you think it is important to have it we're ok adding it.

wietzesuijker:
Thanks @satyaog. Applied all inline suggestions in c25fb90. Added a troubleshooting tip to the quickstart page showing the CLI override.

PR #372, opened 2026-03-30
Add Office Hours header icon

soline-b:


satyaog (commented):
Nice! The link doesn't redirect to the /help page but it looks nice! It's a lot of code juste to add an icon, what do you think? Is the icon worth the code?

overrides/partials/header.html
```diff
@@ -0,0 +1,130 @@
+<!--
```
satyaog, line 3: We could add the following to this file too (like it is in the `footer.html` file). Do you think this is visible enough?

```suggestion
<!-- From https://github.com/squidfunk/mkdocs-material/blob/master/src/templates/partials/header.html,
 with some modifications -->
<!--
```

satyaog (approved):

overrides/partials/header.html
soline-b: Thank you :)
I think it is, but I don't know the good practices for this yet :/

docs/help/index.md
```diff
@@ -79,12 +80,11 @@ We try to provide you different levels of help at Mila, for you to not stay stuc
...
 
-    Office Hours take place in Lab A:
+    {% include-markdown "generated_from_config/office_hours.md" %}
 
-    * from 3pm to 5pm on Tuesdays
-    * from 2pm to 4pm on Wednesdays.
+    They also take place online at the same dates: the virtual invite is on the Mila-core calendar.
 
-    They also take place online at the same dates: the virtual invite is on your Mila calendar.
+    [:octicons-arrow-right-24: Join us at the Office Hours](help/office_hours/)
```
hvdbm: When clicking on the link, it redirects to an invalid page. It seems that the generated URL contains an extra `/help`: `https://mila-docs--372.org.readthedocs.build/en/372/help/help/office_hours/`. 

Based on the link for the FAQ, we could just add `../`:
```suggestion
    [:octicons-arrow-right-24: Join us at the Office Hours](../help/office_hours/)
```

Maybe I'm missing something, but would it work to only use `office_hours/` as a link ? With `../help/` we leave the folder `help` to directly come back in it ?

docs/README.md
```diff
@@ -165,5 +165,6 @@ If you find any errors in the documentation, missing or unclear sections, or wou
     - [Weights and Biases (WandB)](toolbox/wandb.md)
     - [🔗 Research Project Template](https://mila-iqia.github.io/ResearchTemplate)
 - [Get help](help/index.md)
+    - [How to come to the Office Hours](help/office_hours.md)
```
hvdbm: There is a difference between the name of the shortcut "How to come to the Office Hours" and the title of the page "IDT Office Hours". 

Do we want to keep only one name ? Like :
```suggestion
    - [IDT Office Hours](help/office_hours.md)
```

lebrice (approved):

hvdbm (changes requested):
Some comments using the PR preview

docs/help/index.md
soline-b: Nope, it is exactly what is happening. I tried to use this (always redirect to the root) in order to keep a sight on the "deep level" we are on on each page. But I could fix this

docs/README.md
soline-b: Thank you :)

Edit : sorry, I did not see it was suggested changes and just added this in commit `a72d4b5546fe4e3f25815ffcb73985451656e91b` while I was reading your review. But this is really appreciated!

hvdbm (approved):

PR #373, opened 2026-03-31
Fix keyboard shortcuts conflicting with Biel modals

### LINKED ISSUE #367: Search bar disappearing
When searching for something like 'jupyter', the search bar goes away and it goes to a page I didn't search for or select. I am not pressing Enter.

Image

Also, dark mode (a google extension) gets rid of the white window, but maybe people should just avoid using that. 

Image

Also, the old docs seem to just search for whole keywords. 

Image

Image

satyaog:
## Summary
Adds a capture-phase keydown listener that syncs MkDocs Material's hidden `#__search` checkbox with Biel's open-modal state, preventing Material's global shortcuts (e.g. `n`/`p` navigation) from firing while a Biel modal is open.

## Changes
- Added inline `<script>` in `overrides/main.html` that detects open Biel overlays (`biel-search[show-search-modal]`, `biel-bot[show-modal]`) and keeps Material's `#__search` checkbox in sync on every keydown (capture phase, before Material's bubble-phase handler).
- Re-enabled the custom search partial by renaming `overrides/partials/search.html.disabled` → `overrides/partials/search.html`.

## Notes
- The listener uses `.checked` + `dispatchEvent("change")` rather than `.click()` to avoid moving focus away from Biel's modal.
- No continuous `MutationObserver` needed — syncing lazily on keydown is sufficient since Material reads the checkbox synchronously.

resolves #367


satyaog:
This url (https://mila-docs.readthedocs.io/en/staging/) uses the same commit as this PR and can be used to test Biel tools in the doc

soline-b (approved):

lebrice (approved):

PR #382, opened 2026-04-01
Fix search results

### LINKED ISSUE #381: Fix search returning included only files
Searching for `virtual` returns

https://docs.mila.quebec/technical_reference/general_theory/portability/?h=virtual#virtual-environments

as well as

https://docs.mila.quebec/technical_reference/general_theory/python/?h=virtual#virtual-environments

However, `technical_reference/general_theory/python` is a file that is actually only included into `technical_reference/general_theory/portability` and not a page a visitor should land on directly

satyaog:
## Summary
Adds `<!-- START -->` markers and `start=` filters to `include-markdown` directives to prevent front matter and heading duplication when files are transcluded.

## Changes
- Added `<!-- START -->` markers to all included markdown files (acknowledgement, cheatsheet, purpose, singularity, userguides, examples header, etc.) and updated all `include-markdown` calls with `start="<!-- START -->"` to skip front matter.
- Added `search: exclude: true` front matter to files that are only meant to be included (not indexed directly), such as `purpose.md`, `purpose_audience.md`, `python.md`, `portability_modules.md`, singularity partials, and external cluster stubs.

## Notes
- The `search: exclude: true` additions prevent partial/included files from appearing as standalone search results in MkDocs.

resolves #381

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,4 @@
+<!-- START -->
 # Contributing to the Mila Docs
 
 Thank you for your interest into making a better documentation for all at Mila.
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,1 +1,1 @@
-{% include-markdown "../CONTRIBUTING.md" %}
+{% include-markdown "../CONTRIBUTING.md" start="<!-- START -->" %}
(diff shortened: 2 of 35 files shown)


CONTRIBUTING.md
```diff
@@ -1,3 +1,4 @@
+<!-- START -->
```
lebrice: I'm curious, would it be possible to only add these when necessary (when the start flag isnt given, assume the start of the file?)

obilaniu (approved):

soline-b (approved):

PR #399, opened 2026-04-16
Rewrite generated files only if they should be modified

### LINKED ISSUE #395: `on_pre_build` hook causes infinite rebuild loop with `mkdocs serve --livereload`
## Describe the bug

As of #372, running `mkdocs serve --livereload` causes an infinite build loop: the browser page reloads continuously and the terminal shows repeated rebuild triggers. The site never settles into a stable watching state.

## Root cause

`hooks/copy_office_hours.py` registers an `on_pre_build` hook that renders two Jinja2 templates and writes the output directly into `docs/generated_from_config/`:

```
docs/generated_from_config/office_hours.md
docs/generated_from_config/office_hours_short.md
```

MkDocs livereload watches the entire `docs/` directory for file changes. Because the hook writes to that watched directory unconditionally on every build, each build triggers a file-change event, which triggers another build — creating an infinite loop.

## Steps to reproduce

1. Clone the repo and install dependencies
2. Run `mkdocs serve --livereload`
3. Make any change in any watched file (any `.md` should do)
4. Observe the terminal: builds repeat continuously without stopping
5. The browser page reloads in a loop

## Expected behavior

After the initial build, `mkdocs serve` should watch for real content changes and only rebuild when the user edits a source file.

## Proposed fix (by Claude)

Skip writing the file if its content hasn't changed. This avoids touching the file on disk when the output is identical, preventing the spurious change event that triggers the reload loop:

```python
import os

def on_pre_build(config, **kwargs):
    env = Environment(loader=FileSystemLoader("hooks/templates"))
    office_hours_data = config["extra"]["office_hours"]

    for template_file, output_file in [
        ("office_hours_template.md", "office_hours.md"),
        ("office_hours_short_template.md", "office_hours_short.md"),
    ]:
        template = env.get_template(template_file)
        output = template.render(office_hours_data=office_hours_data)
        output_path = f"docs/generated_from_config/{output_file}"

        # Only write if content has actually changed to avoid triggering livereload loop
        if os.path.exists(output_path):
            with open(output_path, "r") as f:
                if f.read() == output:
                    continue

        with open(output_path, "w") as outfile:
            outfile.write(output)
```

**Alternatively**, a more robust fix would be to exclude `docs/generated_from_config/` from the MkDocs livereload watch paths via `watch_exclude` in `mkdocs.yml` (supported in MkDocs ≥ 1.5), though the content-check approach above is simpler and self-contained.

soline-b:
Should fix the issue #395 

resolves #395

### DIFF
--- a/hooks/copy_office_hours.py
+++ b/hooks/copy_office_hours.py
@@ -1,4 +1,5 @@
 from jinja2 import Environment, FileSystemLoader
+import os
 
 
 def on_pre_build(config, **kwargs):
@@ -17,6 +18,12 @@ def on_pre_build(config, **kwargs):
         ]:
         template = env.get_template(template_file)
         output = template.render(office_hours_data=office_hours_data)
+        output_path = f"docs/generated_from_config/{output_file}"
 
-        with open(f"docs/generated_from_config/{output_file}", "w") as outfile:
+        if os.path.exists(output_path):
+            with open(output_path, "r") as infile:
+                if infile.read() == output:
+                    continue # We skip the next steps (writing the file) if it should not be modified
+    
+        with open(output_path, "w") as outfile:
             outfile.write(output)


satyaog (approved):
Works well :)

lebrice (approved):

PR #410, opened 2026-04-27
Delete Threads/Core column and replace with Optimal Bundle column

obilaniu:
What it says on the tin.


soline-b (approved):

lebrice:
Sure. This table is still way too large to be displayed properly, perhaps we should do something about that.

lebrice (approved):

PR #415, opened 2026-05-04
Rework checkpointing example

### LINKED ISSUE #346: misleading preemption handling text and checkpointing example
@obilaniu :

Préemption: https://mila-umontreal.slack.com/archives/CFAS8455H/p1773253182717439 

- Doc sur le sujet un peu périmée => À corriger dans la doc
- Utiliser srun te donne un signal handling correct gratuitement, même pour des jobs à 1 “task” SLURM.

hvdbm:
## Summary
* Add more infos on using checkpoint for model training.
* Add more explanations on `job.sh` and `main.py` code diffs.
   * Remove some typing to reduce the number of unrelated changes to checkpointing. We could rework all the examples to have stronger typing in an other PR.
* Fix a typo in the `get_num_workers` function.

## Issues
* Close #346

### DIFF
--- a/docs/examples/distributed/multi_gpu/main.py
+++ b/docs/examples/distributed/multi_gpu/main.py
@@ -318,7 +318,7 @@ def make_datasets(
 
 
 def get_num_workers() -> int:
-    """Gets the optimal number of DatLoader workers to use in the current job."""
+    """Gets the optimal number of DataLoader workers to use in the current job."""
     if "SLURM_CPUS_PER_TASK" in os.environ:
         return int(os.environ["SLURM_CPUS_PER_TASK"])
     if hasattr(os, "sched_getaffinity"):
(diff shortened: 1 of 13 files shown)


soline-b (approved):

gyom (commented):
I haven't checked in great details the checkpoint saving/loading, but something does jump to my eyes.

We have in the sbatch script `#SBATCH --requeue`, and at the same time we mention in the code later some comment that tells the user that they can requeue the job themselves:
```
    # If you want to directly requeue the job:
    # subprocess.run(["scontrol", "requeue", SLURM_JOBID])
```
I think that this is a good way to inform the user about the way to requeue their job, but at the same time we should mention that they should avoid doing this if they already had `#SBATCH --requeue` earlier. This could be fixed by changing the comment. That is,
```
    # If you want to directly requeue the job:
    # subprocess.run(["scontrol", "requeue", SLURM_JOBID])
    # Be careful to avoid this is you already have the `--requeue` sbatch argument.
```

I asked Claude Sonnet 4.6 and it confirmed what I suspected (for what it's worth).

Otherwise, this PR looks good to me. I'll approve it after this minor line is added to the comments.

gyom (approved):

PR #430, opened 2026-06-05
Standardize section headers in "General Theory"

hvdbm:
## Summary
* Fix duplicated titles in header and in the table of contents.

### Examples
Before :
Capture d’écran, le 2026-06-05 à 15 00 07

After :
Capture d’écran, le 2026-06-05 à 15 00 18


obilaniu (approved):
I vaguely recall that wherever this file is included can also receive a modifier to its include command that says "Shift up or down every header by +- N levels". Could or should that be used instead?

soline-b (approved):

PR #442, opened 2026-06-23
Add Mila logo and colors

soline-b:


satyaog (commented):
I like it a lot! But I find the hovered links too dark and hard to read in dark mode

Image

Maybe we could chat a bit with the communication team to have their quick input? Or try new colours?

soline-b:
> I like it a lot! But I find the hovered links too dark and hard to read in dark mode
> 
> Image
> Maybe we could chat a bit with the communication team to have their quick input? Or try new colours?


I set it slighty lighter than the "no-hover links"? Would it work?

image

satyaog (approved):

PR #454, opened 2026-07-17
Fix broken docs links/anchors and fail the build on regressions

satyaog:
## Summary
Fixes every broken or unrecognized internal link and anchor that MkDocs reports, then turns those problems into build failures so link rot is caught in the PR preview instead of shipping. Also makes canonical URLs correct on Read the Docs preview builds.

## Changes
- Convert directory-style internal links (e.g. `../userguides/wandb`) to their MkDocs `.md` targets across ~30 pages, so they resolve correctly and portably on both docs.mila.quebec and the version-prefixed RTD preview paths.
- Repoint links to the removed `userguides/running_code` page at its replacement, `technical_reference/general_theory/slurm.md`.
- Fix broken anchors: case mismatches (`#Storage`, `#TamIA`, `#Using containers`), the multi-id `Before you begin` heading (split into per-admonition `[](){ #id }` anchors so all three aliases resolve), and the singularity `#nvidia` link now pointing to `nvidia.md`.
- Fix broken image paths in `ai/context.md` and a stale `wandb` anchor link in `gpu_efficiency.md`.
- `mkdocs.yml`: add a `validation.links` block (`not_found`/`anchors`/`unrecognized_links` → `warn`), exclude the obsolete `singularity/README.md`, and resolve `site_url` from `READTHEDOCS_CANONICAL_URL`.
- `.readthedocs.yaml`: set `fail_on_warning: true` so a new broken link/anchor blocks the PR.

## Notes
- Verified with `mkdocs build --strict`: exits 0 with zero warnings.
- Because `fail_on_warning` makes any warning fatal, future changes that trigger unrelated MkDocs warnings will also fail the build — intended, but worth knowing.

### DIFF
--- a/docs/singularity/index.md
+++ b/docs/singularity/index.md
@@ -47,7 +47,7 @@ You can find most of them on `dockerhub`_.
 .. _dockerhub: https://hub.docker.com/
 
 > **tip:**
-        (Optional) You can also pull containers from nvidia cloud see [nvidia](#nvidia)
+        (Optional) You can also pull containers from nvidia cloud see [nvidia](nvidia.md)
 
 Go on `dockerhub`_ and select the container you want to pull.
 
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -28,4 +28,4 @@ build:
 
 mkdocs:
   configuration: mkdocs.yml
-  fail_on_warning: false
+  fail_on_warning: true
--- a/docs/README.md
+++ b/docs/README.md
@@ -6,7 +6,7 @@ description: Technical documentation for Mila's computing infrastructure.
 # Mila technical documentation
 
 Welcome to Mila's technical documentation. If this is your first time here, we
-recommend you start by checking out the [short quick start guide](getting_started/).
+recommend you start by checking out the [short quick start guide](getting_started/index.md).
 
 
 <div class="grid cards" markdown>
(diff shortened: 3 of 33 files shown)


satyaog:
Here is the fail on warnings in action (commit https://github.com/mila-iqia/mila-docs/commit/d54100336f657cb32727438510d330b170d39699) : https://app.readthedocs.org/projects/mila-docs/builds/33635627/

docs/singularity/index.md
```diff
@@ -47,7 +47,7 @@ You can find most of them on `dockerhub`_.
 .. _dockerhub: https://hub.docker.com/
 
 > **tip:**
-        (Optional) You can also pull containers from nvidia cloud see [nvidia](#nvidia)
+        (Optional) You can also pull containers from nvidia cloud see [nvidia](nvidia.md)
```
soline-b, line 50: Are these pages used in the current documentation?

soline-b (approved):

docs/singularity/index.md
satyaog: Ah, maybe not? But why? This could be flagged as an issue and discussed in another PR?

soline-b: Yup, this is not the point of this PR :)

hvdbm (approved):

PR #463, opened 2026-08-05
Add first 'piece of knowledge'

soline-b:
Add an explanation extracted from a Slack interaction.

The following page is added: https://mila-docs--463.org.readthedocs.build/en/463/technical_reference/general_theory/cpu_gpu_numerical_divergence/

### DIFF
--- a/docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
+++ b/docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
@@ -8,7 +8,7 @@ description: >-
 
 ## Summary
 
-Running the *same* model, data, seed, and code on CPU vs. GPU (e.g. via `tensor.to(device)`) will **not** produce bit-identical results. Losses typically match closely at the start of training and then drift apart, eventually diverging completely. This is expected behavior, not a bug — it stems from floating-point arithmetic itself, not from a precision difference between devices.
+Running the *same* model, data, seed, and code on CPU vs. GPU (e.g. via the Pytorch function `tensor.to(device)`) will **not** produce bit-identical results. Losses typically match closely at the start of training and then drift apart, eventually diverging completely. This is expected behavior, not a bug — it stems from floating-point arithmetic itself, not from a precision difference between devices.
 
 ## Root Cause
 
@@ -17,15 +17,16 @@ Running the *same* model, data, seed, and code on CPU vs. GPU (e.g. via `tensor.
 - CPUs and GPUs parallelize reductions (sums, matmuls, etc.) differently to fit their respective hardware, so they add the same set of numbers **in a different order**.
 - Different summation order → different rounding error at each step → tiny discrepancies that compound over many operations and training steps, eventually causing full trajectory divergence (a chaotic, butterfly-effect-style process).
 
-**Takeaway:** never rely on tensor math being bit-exact across devices, or even across different thread counts on the same device.
+!!! tip "Takeaway"
+    Never rely on tensor math being bit-exact across devices, or even across different thread counts on the same device.
 
 ## Why GPUs Are Often (Slightly) More Numerically Stable
 
 - Sequential (naive, one-at-a-time) summation accumulates rounding error linearly with the number of terms.
 - **Pairwise summation** (summing in a tree/divide-and-conquer pattern) accumulates error more slowly — GPUs tend to use pairwise-style reductions to exploit parallelism, which incidentally makes them somewhat better-conditioned at a given precision than a naive sequential CPU sum.
 - This advantage matters most for **ill-conditioned sums**, i.e. when:
-  - the sum of absolute values is much larger than the absolute value of the sum (heavy cancellation), and/or
-  - the ratio between the largest and smallest terms being summed is very large.
+    - the sum of absolute values is much larger than the absolute value of the sum (heavy cancellation), and/or
+    - the ratio between the largest and smallest terms being summed is very large.
 - For well-conditioned sums, the summation order barely matters and CPU vs. GPU differences stay negligible for longer.
 - This is a heuristic tendency, not a guarantee — don't design correctness-critical code around it.
 
--- a/docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
+++ b/docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
@@ -1,0 +1,46 @@
+---
+title: "Why CPU and GPU Training Runs Diverge Numerically"
+description: >-
+  Understand how running the same model on CPU or GPU can lead to different results.
+---
+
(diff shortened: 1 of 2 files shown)


docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
```diff
@@ -0,0 +1,47 @@
+---
+title: "Why CPU and GPU Training Runs Diverge Numerically"
+description: >-
+  Understand how running the same model on CPU or GPU can lead to different results.
```
obilaniu: > how

-> why (?), particulièrement vu l'entête suivant?

docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
```diff
@@ -0,0 +1,47 @@
...
+---
+
+# Why CPU and GPU Training Runs Diverge Numerically
+
+## Summary
+
+Running the *same* model, data, seed, and code on CPU vs. GPU (e.g. via the Pytorch function `tensor.to(device)`) will **not** produce bit-identical results. Losses typically match closely at the start of training and then drift apart, eventually diverging completely. This is expected behavior, not a bug — it stems from floating-point arithmetic itself, not from a precision difference between devices.
```
obilaniu: > it stems from ~floating-point arithmetic itself~ the properties of floating-point arithmetic that make it an imperfect simulation of arithmetic on real numbers.

docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Root Cause
+
+- IEEE float32/float64 are **the same format** on CPU and GPU — there is no inherent precision difference between devices.
```
obilaniu: - IEEE `float32` is **the same format** on both CPU and GPU — there is no inherent precision difference between devices. Likewise, `float64` numbers are interpreted by both CPU and GPU identically.

(Avoid a possible misread that the FP32 format is the same as FP64).

docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Why GPUs Are Often (Slightly) More Numerically Stable
+
+- Sequential (naive, one-at-a-time) summation accumulates rounding error linearly with the number of terms.
+- **Pairwise summation** (summing in a tree/divide-and-conquer pattern) accumulates error more slowly — GPUs tend to use pairwise-style reductions to exploit parallelism, which incidentally makes them somewhat better-conditioned at a given precision than a naive sequential CPU sum.
+- This advantage matters most for **ill-conditioned sums**, i.e. when:
+    - the sum of absolute values is much larger than the absolute value of the sum (heavy cancellation), and/or
+    - the ratio between the largest and smallest terms being summed is very large.
+- For well-conditioned sums, the summation order barely matters and CPU vs. GPU differences stay negligible for longer.
+- This is a heuristic tendency, not a guarantee — don't design correctness-critical code around it.
```
obilaniu: I know I wrote this on the Slack thread this was sourced from, but I think that may have been unnecessary editorializing. Also, on CPU some libraries (or languages, like Python as of 3.x for some x) have started using compensated summation algorithms to erase the disadvantages of naïve summation.

I would scrap this entirely, it's part I said that is most heavily salted with unspoken conditionals and assumptions. All else equal a GPU doing pairwise summation will calculate more accurate sums than a CPU doing a linear accumulation, but all else is rarely equal and CPU implementations are rarely this naïve.

docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Mitigation Strategies (and their costs)
+
+| Strategy | Effect | Cost / Caveat |
+|---|---|---|
+| Use `float64` instead of `float32` | Reduces rounding error per operation, often "fixes" the visible divergence in practice | On **CPU**: reasonable — x86 does float64 addition at roughly half the throughput of float32. On **GPU**: usually a bad idea — most consumer/data-center GPUs have float64 throughput at only 1/24, 1/32, or even 1/64 of float32 (only workstation-grade cards like some Teslas get closer to 1/2). |
+| Sort terms before summing (smallest → largest), or use a priority queue for streaming sums | Improves conditioning of sequential summation, closing some of the gap with pairwise summation | Extra sort/heap overhead; only helps when the sum is ill-conditioned to begin with. |
```
obilaniu: I would delete the line about sorting or priority queue entirely. A Kahan summation algorithm gets you a much more solid approximation in linear time, even if it is still subject to rounding errors and order. And you should not implement Kahan yourself - you should do what the next line says and use `np.sum()`, `torch.sum()`, ... and rely on libraries to do it for you.

docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
```diff
@@ -0,0 +1,47 @@
...
+| Use `torch.sum()`'s precision argument | Lets you request higher internal accumulation precision without changing the whole model's dtype | Still subject to the general CPU/GPU order-of-operations caveat above. |
```
obilaniu: > Still subject to the general CPU/GPU order-of-operations caveat above, but much less sensitive to it.

docs/technical_reference/general_theory/cpu_gpu_numerical_divergence.md
```diff
@@ -0,0 +1,47 @@
...
+| Accept the divergence | Often the right call for training runs where exact reproducibility isn't required | Only track/compare aggregate metrics (final loss, accuracy) across devices rather than exact per-step trajectories. |
+
+## Practical Guidance
+
+1. If you see losses that match early in training and drift apart later when switching CPU ↔ GPU, this is very likely just accumulated floating-point rounding — check datatype, seeds, and code parity as a sanity check, but don't expect them to fix it.
+2. Don't chase bit-exact reproducibility across devices or thread counts as a goal in itself; instead, define what "close enough" means for your use case (e.g. metrics within some tolerance).
+3. If you need better numerical stability and can afford the cost, try `float64` **on CPU** first — it's the cheapest reliable fix. Avoid `float64` on GPU unless you're on stability-oriented hardware, since the throughput penalty is severe.
+4. If you suspect ill-conditioned sums specifically (large dynamic range or heavy cancellation among terms), sorting before summation or using a priority queue can help without paying the full float64 cost.
```
obilaniu: Delete for reasons explained above. Should be a rather uncommon situation in ML and would indicate other problems anyways.

obilaniu (changes requested):

PR #477, opened 2026-09-04
Update DocSearch to v5

soline-b:
* Update DocSearch according to https://docsearch.algolia.com/docs/migrating-from-v4/
* Minor changes in search bar CSS

### DIFF
--- a/overrides/main.html
+++ b/overrides/main.html
@@ -15,12 +15,12 @@ and we'll take a look at it!
 <meta name="algolia-site-verification" content="3EAD36F59F913078" />
 
 <!-- DocSearch styles -->
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@4" />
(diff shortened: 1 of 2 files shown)


overrides/main.html
```diff
@@ -81,18 +81,13 @@
...
-        indices: [
-            {
-                name: "MILA Docs",
-            },
-        ],
-        assistantId: "d9b031b9-047c-4b63-a794-010d037dce19",
-        agentStudio: true,
+        //indices: ["MILA-Docs"],
+        agentId: "d9b031b9-047c-4b63-a794-010d037dce19",
         // suggestedQuestions: true,
```
hvdbm: Do we still need the commented fields `indices` and `suggestedQuestions` ?

hvdbm (approved):

obilaniu (approved):

overrides/main.html
soline-b: This is not needed, I'll delete them :)

